PR #185, opened 2017-11-10
Fix broken editor rendering after changing font size on guest portals

### LINKED ISSUE #174: Guest editors render incorrectly after guest changes font size and host switches tabs
Steps to reproduce:

1. Open 2 files on host
2. As guest, change font size
3. Switch to second file on host
4. On guest, observe something like the following:

screen shot 2017-11-03 at 4 06 30 pm

screen shot 2017-11-03 at 4 05 34 pm

as-cii:
Fixes #174 

When editor styles change, Atom notifies only editor components that are attached to the DOM at the moment of the change. Thus, if the element we are about to add had already been rendered, we will preemptively clear all its styling information to ensure it will render correctly even if some styles changed while it was not attached to the DOM.

This is admittedly something that we should have better support for in core, but this workaround should be an okay stopgap measure.

/cc: @nathansobo @jasonrudolph

### DIFF
--- a/lib/guest-portal-binding.js
+++ b/lib/guest-portal-binding.js
@@ -155,6 +155,8 @@ class GuestPortalBinding {
     this.newActivePaneItem = newActivePaneItem
 
     if (this.activePaneItem) {
+      this.ensurePaneItemStylesAreUpToDate(newActivePaneItem)
+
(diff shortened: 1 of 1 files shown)


oz90210:
hello

PR #198, opened 2017-11-15
Trim whitespace from portal ID

50Wliu:
### Requirements

* Filling out the template is required. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
* All new code requires tests to ensure against regressions.

### Description of the Change

Trims any whitespace off of a potential portal ID.  The automatic clipboard reading is super cool :).

### Alternate Designs

None.

### Benefits

Better portal ID validation.

### Possible Drawbacks

None?

### Applicable Issues

Fixes #197.

PR #205, opened 2017-11-16
Streamline tests related to fetching portal ID from clipboard

jasonrudolph:
The changes in https://github.com/atom/teletype/pull/198 duplicated the test for how we handle the scenario where the clipboard contains content that is _not_ a portal ID:

https://github.com/atom/teletype/blob/5a4a7d473265396c14ad9a0ce435b3529040fb13/test/portal-list-component.test.js#L180-L184

Those assertions ☝️ existed prior to #198, and #198 duplicates those existing assertions:

https://github.com/atom/teletype/blob/5a4a7d473265396c14ad9a0ce435b3529040fb13/test/portal-list-component.test.js#L197-L201

This PR removes that duplication. 

---

Also, prior to #198, we had a single `test` function that covered all scenarios related to fetching the portal ID from the clipboard. #198 added a second `test` function specifically related to the scenario where the clipboard contains a portal ID surrounded by whitespace. This PR consolidates that scenario into the pre-existing `test` function, along with the other related scenarios.

### DIFF
--- a/test/portal-list-component.test.js
+++ b/test/portal-list-component.test.js
@@ -172,28 +172,20 @@ suite('PortalListComponent', function () {
     const {clipboard} = component.props
     const {joinPortalComponent} = component.refs
 
+    // Clipboard containing a portal ID
     clipboard.write('bc282ad8-7643-42cb-80ca-c243771a618f')
     await joinPortalComponent.showPrompt()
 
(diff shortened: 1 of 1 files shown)


as-cii:
👍

50Wliu:
:heart:

PR #212, opened 2017-11-16
Show initialization failures also in the popover

### LINKED ISSUE #210: Uncaught TypeError: Cannot read property 'onDidChange' of null
[Enter steps to reproduce:]

1. ...
2. ...

**Atom**: 1.22.1 x64
**Electron**: 1.6.15
**OS**: Mac OS X 10.13.1
**Thrown From**: [teletype](https://github.com/atom/teletype) package 0.1.2

### Stack Trace

Uncaught TypeError: Cannot read property 'onDidChange' of null

```
At /Users/manu_hdez/.atom/packages/teletype/lib/portal-list-component.js:14

TypeError: Cannot read property 'onDidChange' of null
    at /packages/teletype/lib/portal-list-component.js:14:57)
    at render (/packages/teletype/node_modules/etch/lib/render.js:17:25)
    at updateChildren (/packages/teletype/node_modules/etch/lib/patch.js:108:36)
    at patch (/packages/teletype/node_modules/etch/lib/patch.js:18:9)
    at updateSync (/packages/teletype/node_modules/etch/lib/component-helpers.js:116:20)
    at /packages/teletype/node_modules/etch/lib/component-helpers.js:72:7
    at DefaultScheduler.performUpdates (/packages/teletype/node_modules/etch/lib/default-scheduler.js:48:34)
```

### Commands

```
```

### Non-Core Packages

```
autoprefixer 3.7.1 
emmet 2.4.3 
pigments undefined 
sass-autocompile 0.13.3 
teletype 0.1.2 
```

as-cii:
Fixes #210 

Previously, we would only show a notification when failing to initialize the client (e.g. this could happen when an error occurs while checking the protocol version) without informing the popover. This was also causing exceptions like #210, because we were failing to instantiate objects like `PortalBindingManager` due to the initialization error.

With this pull-request we will:

* Show an error message in the popover too, coloring the status bar indicator as red when that happens.
* Don't attempt to re-initialize the client if it is outdated or an error has previously occurred.

<p align=center>

</p>

I have tested this works correctly with a few themes (One Light UI, One Dark UI, Atom Material, Nord UI) by opening the DevTools and enabling the `Offline` checkbox in the Network tab. On reload, I verified that the above message was being shown in the popover.

/cc: @jasonrudolph @nathansobo

### DIFF
--- a/styles/teletype.less
+++ b/styles/teletype.less
@@ -411,13 +411,7 @@
   }
 }
 
-.PackageOutdatedComponent {
-  color: @text-color;
-  text-align: center;
-  padding: 0 @component-padding @component-padding @component-padding;
-}
-
-.PackageInitializationErrorComponent {
+.PackageOutdatedComponent, .PackageInitializationErrorComponent {
   color: @text-color;
   text-align: center;
   padding: 0 @component-padding @component-padding @component-padding;
--- a/styles/teletype.less
+++ b/styles/teletype.less
@@ -416,3 +416,9 @@
   text-align: center;
   padding: 0 @component-padding @component-padding @component-padding;
 }
+
+.PackageInitializationErrorComponent {
+  color: @text-color;
+  text-align: center;
+  padding: 0 @component-padding @component-padding @component-padding;
+}
--- a/styles/teletype.less
+++ b/styles/teletype.less
@@ -18,7 +18,7 @@
     }
   }
 
-  &.outdated {
+  &.outdated, &.initialization-error {
     color: @text-color-error;
 
     &:hover {
--- a/lib/package-initialization-error-component.js
+++ b/lib/package-initialization-error-component.js
@@ -1,0 +1,31 @@
+const etch = require('etch')
+const $ = etch.dom
+
+module.exports =
+class PackageInitializationErrorComponent {
+  constructor (props) {
+    this.props = props
+    etch.initialize(this)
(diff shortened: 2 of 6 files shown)


styles/teletype.less
```diff
@@ -416,3 +416,9 @@
   text-align: center;
   padding: 0 @component-padding @component-padding @component-padding;
 }
+
+.PackageInitializationErrorComponent {
+  color: @text-color;
+  text-align: center;
+  padding: 0 @component-padding @component-padding @component-padding;
+}
```
jasonrudolph: Since these are the same styles used for `.PackageOutdatedComponent`, should we get rid of the duplication?

```less
.PackageOutdatedComponent, .PackageInitializationErrorComponent {
  color: @text-color;
  text-align: center;
  padding: 0 @component-padding @component-padding @component-padding;
}
```

as-cii: Good point, I will address that shortly.

as-cii:
> Can I talk you into describing the steps you used to test these changes?

Thanks for reminding me of that, I have updated the body of the pull request to describe the steps I followed.

I have also addressed https://github.com/atom/teletype/pull/212#discussion_r151428811 and verified it works as expected, so I am going to merge this once CI is green. Thanks for the review, @jasonrudolph! 🙇

PR #231, opened 2017-11-17
Add RFC: Allow guests to view multiple remote buffers

nathansobo:
This is an experiment in a more transparent approach to decision-making on major product features. The idea is to create "request for comment" style posts clearly outlining major changes. It is heavily based on the [approach used by the Rust community](https://github.com/rust-lang/rfcs), but I've adapted it a bit since we're not building a programming language.

We can experiment with a couple of these and then add some documentation to our contribution guide to help community members create them as well.

@as-cii I know we already talked about this, so if you want to get started implementing it before we get this RFC perfect, I think that would be fine. I'd just like to have this document in nice shape before we merge anything so we can get feedback from the community.

/cc @as-cii @jasonrudolph

### DIFF
--- a/doc/rfcs/001-allow-guests-to-open-multiple-remote-buffers.md
+++ b/doc/rfcs/001-allow-guests-to-open-multiple-remote-buffers.md
@@ -34,13 +34,13 @@ Editors for remote buffers are *only* automatically opened when you are followin
 
 As a guest of a shared workspace, you don't automatically receive access to the host's file system. Tools like the tree view, the file finder, and project search don't have access to their current project. (This capability will be the subject of an upcoming RFC).
 
-For now, however, the file finder *will* be augmented to make it easy for you to navigate any buffer that the host current has open in their workspace. These remote buffers will be decorated with the avatar of the host that owns their workspace.
+For now, however, the file finder *will* be augmented to make it easy for you to navigate any buffer that the host currently has open in their workspace. These remote buffers will be decorated with the avatar of the host that owns their workspace.
 
 ## Drawbacks
 
 Currently, a portal always corresponds to a single tab. This keeps things really simple. To leave the portal, you just close the tab. Now there could be multiple editors in your workspace that are part of a remote workspace, which increases complexity.
 
-When you follow someone between multiple buffers, editors for remote buffers may start to stack up in your workspace. Some may consider this to be too cluttered. If this becomes problematic, we could adding an option to automatically close an editor upon following a collaborator to a different buffer or just make that the default behavior.
+When you follow someone between multiple buffers, editors for remote buffers may start to stack up in your workspace. Some may consider this to be too cluttered. If this becomes problematic, we could add an option to automatically close an editor upon following a collaborator to a different buffer or just make that the default behavior.
 
 ## Rationale and alternatives
 
--- a/doc/rfcs/001-allow-guests-to-open-multiple-remote-buffers.md
+++ b/doc/rfcs/001-allow-guests-to-open-multiple-remote-buffers.md
@@ -50,7 +50,7 @@ When you follow someone between multiple buffers, editors for remote buffers may
 
 ## Rationale and alternatives
 
-The workspace and the project are *both* pieces of state worth sharing. The workspace is more transient, representing a "working set" of resources that a developer is looking at *right now*. Ideally, anything in resource in this working set should be able to be accessed by collaborators. This includes the set of open buffers, but it could also include other state such as a terminal session, a debugger, console output, etc.
+The workspace and the project are *both* pieces of state worth sharing. The workspace is more transient, representing a "working set" of resources that a developer is looking at *right now*. Ideally, any resource in this working set should be able to be accessed by collaborators. This includes the set of open buffers, but it could also include other state such as a terminal session, a debugger, console output, etc.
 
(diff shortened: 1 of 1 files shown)


doc/rfcs/001-allow-guests-to-open-multiple-remote-buffers.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Following
+
+When you first join a portal as a guest, you're automatically *following* the host. This means that you jump to the host's current position and continue to follow them as they move between buffers.
+
+In order to jump to the host's current position, an editor is automatically added to your workspace for their active buffer when you first join their workspace. If you continue to follow the host, any time they switch to a new buffer, a new editor for that remote buffer is automatically added to your workspace and focused. Existing editors for previous remote buffers are not automatically closed when this switch occurs.
+
+If you stop following the host, either manually or automatically due to them navigating away while you are still typing, then when they change their active editor your workspace is unaffected. If you start following the host again then a new editor may be added to your workspace based on their current location so that you can continue to follow them.
```
as-cii: > If you stop following the host, either manually or automatically due to them navigating away while you are still typing, then when they change their active editor your workspace is unaffected

So essentially what this means is that tabs are opened automatically only when you are following. If you are not, we will not open tabs for new buffers, but you can still access them via the file finder. Am I interpreting this correctly? If so, I really like this behavior! Do you think we could clarify such aspect a little more in this paragraph?

doc/rfcs/001-allow-guests-to-open-multiple-remote-buffers.md
```diff
@@ -0,0 +1,47 @@
...
+
+You can follow any other guest participating in the host's workspace in the exact same way. If they move between buffers, you will follow them. The host does not enjoy any special privilege with respect to the ability to be followed between different files.
```
as-cii: Love this. Basically the only privilege hosts have is to decide which files are shared by opening them in their workspace.

doc/rfcs/001-allow-guests-to-open-multiple-remote-buffers.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Navigation
+
+As a guest of a shared workspace, you don't automatically receive access to the host's file system. Tools like the tree view, the file finder, and project search don't have access to their current project. (This capability will be the subject of an upcoming RFC).
+
+For now, however, the file finder *will* be augmented to make it easy for you to navigate any buffer that the host current has open in their workspace. These remote buffers will be decorated with the avatar of the host that owns their workspace.
```
as-cii: In addition to the avatar, I think the file finder could use the same format as the one we currently use for remote tabs, e.g. `[avatar] @nathansobo: some-file.js`

doc/rfcs/001-allow-guests-to-open-multiple-remote-buffers.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Drawbacks
+
+Currently, a portal always corresponds to a single tab. This keeps things really simple. To leave the portal, you just close the tab. Now there could be multiple editors in your workspace that are part of a remote workspace, which increases complexity.
```
as-cii: A low-effort idea to solve this would be to add back the `Teletype: Leave Portal` command. It'd be symmetric with the `Teletype: Close Portal` command we currently offer to hosts and would be pretty easy to access too (we could even provide default keybindings for it).

doc/rfcs/001-allow-guests-to-open-multiple-remote-buffers.md
nathansobo: Yeah, that was my intent. I can clarify this.

doc/rfcs/001-allow-guests-to-open-multiple-remote-buffers.md
```diff
@@ -0,0 +1,55 @@
...
+
+You can follow any other guest participating in the host's workspace in the exact same way. If they move between buffers, you will follow them. The host does not enjoy any special privilege with respect to the ability to be followed between different files.
+
+Editors for remote buffers are *only* automatically opened when you are following another collaborator. If you are not following someone, no editors are automatically opened. When you start following another collaborator again, an editor will be automatically opened based on their location. You can also open any buffer in the host's workspace directly by navigating to it...
+
+### Navigation
+
+As a guest of a shared workspace, you don't automatically receive access to the host's file system. Tools like the tree view, the file finder, and project search don't have access to their current project. (This capability will be the subject of an upcoming RFC).
+
+For now, however, the file finder *will* be augmented to make it easy for you to navigate any buffer that the host currently has open in their workspace. These remote buffers will be decorated with the avatar of the host that owns their workspace.
```
jasonrudolph: > These remote buffers will be decorated with the avatar of the host that owns their workspace.

Nice. Do you have any particular type of decoration in mind?

Arcanemagus: Will custom icon packages be prevented from overriding this avatar? I can see it being problematic if this isn't prevented and these remote tabs aren't yet separated.

jasonrudolph: Thanks for the feedback, @Arcanemagus. 

We're opting to let the RFC remain vague regarding the specific UI design details so that we can explore various options. However, the changes should be additive in nature (i.e., not removing/replacing any existing UI components).

jasonrudolph:
@as-cii and I have updated the RFC to propose answers to the questions asked in https://github.com/atom/teletype/pull/231#pullrequestreview-77489844. We'll leave this pull request open for a bit longer to give folks time to share any additional feedback.

as-cii:
We're actively working on this proposal, so I went ahead and merged it into master.

jasonrudolph:
FYI: We're tracking the progress of this RFC in https://github.com/atom/teletype/issues/268. 📝

Issue #47, opened 2017-09-15
Provide better error message when attempting to join portal with empty portal ID


jasonrudolph:
### How to reproduce

1. Trigger "Join Portal" command
2. When prompted for a portal ID, hit "enter" without providing a portal ID
3. Observe error message

### Expected behavior

If the user doesn't provide a portal ID, we shouldn't even attempt to join the portal. Instead, we should tell the user that they need to provide a valid portal ID.

### Demo

![demo](https://user-images.githubusercontent.com/2988/30496821-a0174be2-9a1e-11e7-843a-ae6f0669ffb3.gif)

Issue #131, opened 2017-10-25
Button should be disabled if text field is empty


ungb:
This is minor but can be a usability issue as I don't think the scenarios where the text field is empty are valid.  Buttons should be disabled when field is empty.

Enter your Token:
![image](https://user-images.githubusercontent.com/5624870/32023087-20d13e40-b98d-11e7-8f91-cb9be29c1d50.png)

Joining Portal:

![image](https://user-images.githubusercontent.com/5624870/32023127-41ae1d40-b98d-11e7-8a82-d555d6ebed73.png)

jasonrudolph:
Closed by https://github.com/atom/teletype/pull/248. These improvements will go out in the next release.

Ben3eeE:
@jasonrudolph It looks like you intended to close this issue but clicked the wrong button :slightly_smiling_face: Please reopen I am misunderstanding something!

PR #243, opened 2017-11-20
Disable join button if portal ID is invalid.

tvand7093:
### Requirements

### Description of the Change

When attempting to join a new session, the join button did no validation on the specified portal id. This change now disables the join button in the following scenarios:
1. When the portal id is blank
2. When the portal id is not a valid UUID

Once the value has been validated, the button becomes enabled and allows the user to click it.

### Alternate Designs

Originally I considered just showing an error message when the button was clicked, but this felt sloppy since this is more of a input validation step. Using this method, the button prevents the attempt all together unless it is valid. I did include some guards in the join step to prompt the user in the event that an invalid portal id was somehow specified/allowed. I wasn't fully sure if this was something that we would want, so I left in just in case and for some extra protection. I can remove it if requested.

### Benefits

Users will not get a cryptic message when attempting to connect with an invalid id value. It also feels more like a traditional form input since the button blocks all invalid inputs now.

### Possible Drawbacks

Since the button automatically disables/enables, it is possible that somebody may not understand why the button is disabled in the event of an invalid UUID.

### Applicable Issues

- #47

### DIFF
--- a/lib/join-portal-component.js
+++ b/lib/join-portal-component.js
@@ -54,7 +54,7 @@ class JoinPortalComponent {
   async validatePortalId() {
     //here we validate the portal id is correct. 
     const portalId = this.refs.portalIdEditor.getText().trim()
-    await this.update({ isValidPortalId: portalId && isUUID(portalId) })
+    await this.update({ isValidPortalId: isUUID(portalId) })
   }
 
   async showPrompt () {
--- a/lib/join-portal-component.js
+++ b/lib/join-portal-component.js
@@ -11,7 +11,7 @@ class JoinPortalComponent {
       'core:confirm': this.joinPortal.bind(this),
       'core:cancel': this.hidePrompt.bind(this)
     })
-  }
+}
 
   destroy () {
     this.disposables.dispose()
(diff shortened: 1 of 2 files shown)


tvand7093:
Is there a way to see why exactly my build is failing? I see that it's running the tests, but I can't see any output so I don't know exactly what is going on.

jasonrudolph:
Thanks for your contribution, @tvand7093! :bow: Building on your pull request, we [added some commits](https://github.com/atom/teletype/compare/f0072fd7c583c218e41fd29fa640242d7e32f72f...1bdb2da2f83eeb0d8ce65ec9bf6b1513cc367cae) and merged your branch. If you're interested, each commit message explains the change and the motivation for that change.

These improvements will go out in the next release of the package, likely later this week.

tvand7093:
I wasn’t aware of the readAfterUpdate override! That makes it way cleaner. Just a heads up, I’ll likely be taking on #131 over the next few days since it’s similar to this one.

as-cii:
> Just a heads up, I’ll likely be taking on #131 over the next few days since it’s similar to this one.

That would be great, thanks @tvand7093! ⚡️

jasonrudolph:
> These improvements will go out in the next release of the package, likely later this week.

teletype v0.2.2 is now available, and it includes these improvements. :zap:

PR #248, opened 2017-11-21
Sign-in button is now disabled when token is empty or whitespace

tvand7093:
### Description of the Change
When first signing in, the sign in button is enabled when there is no value in the token field. This changes makes it so that the sign in button is disabled when:
1. The token field is empty
2. The token is all whitespace

Once a valid character is typed into the field, the button becomes enabled. This also includes an error prompt in the event someone was able to click the button when the token is empty.

### Alternate Designs
These change follow similar changes found [here](https://github.com/atom/teletype/compare/f0072fd7c583c218e41fd29fa640242d7e32f72f...1bdb2da2f83eeb0d8ce65ec9bf6b1513cc367cae). For this reason, the code is rather similar and not many other decisions were made.

### Benefits
This prevents the user from signing in with an empty token, which is more of a standard practice form validation practice.

### Possible Drawbacks
N/A

### Applicable Issues
#131


test/teletype-package.test.js
```diff
@@ -955,6 +955,88 @@ suite('TeletypePackage', function () {
     assert(description.includes('some error'))
   })
 
+  test('disables button when empty token specified', async () => {
+    {
+        const env = buildAtomEnvironment()
```
jasonrudolph: The indentation here appears to be inconsistent with the rest of the code in the repository. Can you please update it to use 2 spaces for indentation. 

For style guidelines, please see https://standardjs.com/#the-rules and [CONTRIBUTING.md#javascript-styleguide](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#javascript-styleguide).

tvand7093: I was following the pattern as seen in tests [above](https://github.com/atom/teletype/blob/master/test/teletype-package.test.js#L881-L936) where there are multiple cases for one `test` method.

lib/sign-in-component.js
```diff
@@ -14,6 +14,9 @@ class SignInComponent {
     this.disposables.add(this.props.commandRegistry.add(this.element, {
       'core:confirm': this.signIn.bind(this)
     }))
+    this.props.workspace.observeTextEditors(editor => {
+      this.attachChangeEvent()
```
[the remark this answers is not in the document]
tvand7093: It is needed there so that way when the dialog first comes up, the editor is wired up to the event. If this code is commented out, then the button will never be able to change state. I have a test case that demonstrates this. If there is a more appropriate way to handle this, I'm all ears and will address it ASAP!

tvand7093:
@jasonrudolph as for the build, the only thing I can think of is that I'm doing this on a Mac, so perhaps there is something strange secretly going on...

tvand7093:
After looking at a diff between the two build logs is the `TEST_TIMEOUT_IN_MS` env variable is being set in your build and not mine. Perhaps that is causing some issues? 🤔

jasonrudolph:
> After looking at a diff between the two build logs is the `TEST_TIMEOUT_IN_MS` env variable is being set in your build and not mine. Perhaps that is causing some issues? 

Nice catch. That was indeed the problem. The build is green now. 😅 

I'll circle back around to take another look at this pull request :soon:.

lib/sign-in-component.js
```diff
@@ -70,10 +78,31 @@ class SignInComponent {
...
+        this.refs.loginButton.disabled = !token
+      })
+    }
+  }
+
   async signIn () {
+    const token = this.refs.editor.getText().trim()
+    
+    if (!token) {
+      this.props.notificationManager.addError('Invalid login token', {
```
jasonrudolph: Can we use `renderErrorMessage` here instead of a notification?

If we use a notification, that notification will appear all the way in the upper-right corner of the app. If we use `renderErrorMessage`, we can show the error message right next to the token field (i.e., right next to the error 🌟 ).

tvand7093:
@jasonrudolph thanks for all the feedback! I can for sure take care of these changes. I will probably get to them over the next few evenings (aside from thanksgiving)! :turkey:

lib/sign-in-component.js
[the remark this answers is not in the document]
tvand7093: I’m still new to using etch how so I wasn’t aware of when exactly the refs were available. Learn something new every day!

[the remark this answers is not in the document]
jasonrudolph: You and me both! I just learned this today, too. 😸

lib/sign-in-component.js
```diff
@@ -31,6 +30,13 @@ class SignInComponent {
     return etch.update(this)
   }
 
+  readAfterUpdate () {
```
tvand7093: Because the view refreshes, we have to rewire the `onDidChange` event otherwise the sign in button will be stuck in a disabled state, even when a token value is input.

test/teletype-package.test.js
```diff
@@ -988,6 +988,29 @@ suite('TeletypePackage', function () {
...
+    }
+
+    const {popoverComponent} = pack.portalStatusBarIndicator
+    const {signInComponent} = popoverComponent.refs
+
+    signInComponent.refs.editor.setText('some-token')
+
+    await signInComponent.signIn()
+
+    // TODO: Find out why these refs are not updating.
```
tvand7093: @jasonrudolph: This test case is supposed to test what happens when a token value is provided, the user clicks sign in, and the sign in fails. Testing using atom, it resets the token field and disables the button. However, it seems that I cannot get the refs here to be updated to actually test how it worked. I have banged my head on this all weekend... Any ideas?

jasonrudolph:
@tvand7093: Thanks for the updates. I've got this queued up for review. I'll follow up once I've reviewed the latest updates. :bow:

jasonrudolph:
@tvand7093: Just a quick note to let you know that I haven't forgotten about you. We're wrapping up some big updates, and then I'll circle back around to this pull request.

jasonrudolph:
@tvand7093: I pushed up a few updates, mostly related to preventing the need for resubscribing to text editor change events each time the popover is re-rendered. :smile:

Thanks again for this contribution! :bow::beers:

PR #252, opened 2017-11-21
👕  Set up Standard JS linter

jasonrudolph:
Configure linter for [JavaScript Standard Style](https://standardjs.com) to foster consistency with the [Atom style guide](https://github.com/atom/atom/blob/572aec74d9fb612849a2b02cb5e54d515341c0fd/CONTRIBUTING.md#javascript-styleguide).

### TODO

- [x] Set up [JavaScript Standard Style](https://standardjs.com) linter
- [x] Resolve existing linter violations

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
   "main": "index.js",
   "repository": "https://github.com/atom/teletype",
   "scripts": {
+    "lint": "standard --verbose",
     "test": "atom --test test"
   },
   "author": "",
--- a/package.json
+++ b/package.json
@@ -66,7 +66,8 @@
       "mocha": true
     },
     "globals": [
-      "atom"
+      "atom",
+      "ErrorEvent"
     ],
     "ignore": ["test/fixtures"]
   }
--- a/package.json
+++ b/package.json
@@ -62,14 +62,11 @@
     }
   },
   "standard": {
+    "env": {
+      "mocha": true
+    },
     "globals": [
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -58,5 +60,17 @@
       "default": "mt1",
       "order": 3
     }
+  },
+  "standard": {
+    "globals": [
+      "atom",
+      "setup",
```
50Wliu: Consider adding the `mocha` env instead of specifying all the mocha globals?

jasonrudolph: @50Wliu: That sounds promising. :smile: Can you point me toward an example?

Arcanemagus: Exactly like is done [here](https://github.com/atom/atom/blob/572aec74d9fb612849a2b02cb5e54d515341c0fd/package.json#L178-L189), but with `"mocha": true`.

Note that `mocha` is a [predefined environment](https://eslint.org/docs/user-guide/configuring#specifying-environments).

as-cii: Nice, yeah I think we should use the `mocha: true` environment.

jasonrudolph: Thanks! db56166f6510e04df6b541642c3c108b2da21fd8 adds the `"mocha": true` env config.

jasonrudolph:
@as-cii: Thanks for the review. I pushed up several commits to address the linter violations in the `test` dir. Would you mind reviewing those changes?

PR #270, opened 2017-12-12
Save buffer on the host when attempting to save it on the guest

as-cii:
Closes https://github.com/atom/teletype/issues/240

This pull request uses the APIs introduced in https://github.com/atom/teletype-client/pull/45 to:

* Perform a save request when attempting to save a remote buffer on the guest (e5aa4c0).
* Honor the save request on the host by flushing changes to disk. Requests to save in-memory buffers are ignored by the host. (0edbe49)

### Verification

* Open a new Atom instance and share a portal.
* Open a new Atom instance and join the previously shared portal.
* Edit the buffer on the guest portal and save.
* Observe saves being relayed to the host buffer.

/cc: @nathansobo @jasonrudolph

PR #272, opened 2017-12-12
Remove duplicate logic for calculating remote prefix

jasonrudolph:
In #270, 3a30144db33bf2858f347a4e112e59228cc030e6 duplicates the logic for determining the prefix for a remote title/path.

https://github.com/atom/teletype/blob/3a30144db33bf2858f347a4e112e59228cc030e6/lib/buffer-binding.js#L48-L49

https://github.com/atom/teletype/blob/3a30144db33bf2858f347a4e112e59228cc030e6/lib/editor-binding.js#L65-L66

This pull request proposes a refactoring that eliminates that duplication and removes the dependency on `Portal` in `EditorBinding` and `BufferBinding`.

@as-cii: What do you think?

### DIFF
--- a/lib/buffer-binding.js
+++ b/lib/buffer-binding.js
@@ -5,11 +5,11 @@ function doNothing () {}
 
 module.exports =
 class BufferBinding {
-  constructor ({buffer, portal, isHost, didDispose}) {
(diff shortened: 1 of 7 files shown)


as-cii:
Nice work, thanks!

PR #279, opened 2017-12-13
Close/leave portals when deactivating TeletypePackage

as-cii:
Previously, deactivating the TeletypePackage would only dispose the bindings without tearing down the associated portal instances. This would cause the disposed bindings to keep receiving messages from e.g. Portal or EditorProxy and would throw errors because the binding methods were called while being in a inconsistent/disposed state.

This pull request addresses the exceptions we were observing in the DevTools when running tests locally by closing/leaving all the portals managed by PortalBindingManager when tearing down the package.

/cc: @jasonrudolph


jasonrudolph:
Thanks for tracking down the cause of these exceptions. :zap:

PR #283, opened 2017-12-14
Show site position avatars inside remote pane items

### LINKED ISSUE #276: Avatar isn't in the correct pane in split pane view.
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:

### Description

Avatar isn't in the correct pane in split view.

### Steps to Reproduce

1. Share a portal on Atom.
1. Guest should be following while host moves 1 buffer to the bottom.

**Expected behavior:**

Avatar is in the middle of the active buffer

**Actual behavior:**

Avatar shows up on middle of atom 
image

NOTE: if you split right it will show up in the right pane only and never the left one.
image

**Reproduces how often:**
100%

### Versions

commit 9b8a3cde53e246184ba260034a80ca750627a6dd

as-cii:
Fixes #276

Previously, we would show site position avatars on the right hand side of the workspace center element. As illustrated in the aforementioned issue, this could cause a confusing behavior when interacting with split panes:

screen shot 2017-12-14 at 11 27 22

With this pull request we are moving site position avatars from the "workspace center" element to the currently active remote pane item:

| before  | after   |
|---------|---------|
| before | after |

### Verification steps

On a portal with one or more guests:

* Create some split items on the host (or, analogously, move remote buffers into other panes on the guest).
* Move among tabs and notice how site position avatars follow the currently active remote pane item.
* Notice also how focusing a non-remote pane item (e.g. git diff view) hides the avatars.

/cc: @jasonrudolph @ungb @rsese

### DIFF
--- a/lib/guest-portal-binding.js
+++ b/lib/guest-portal-binding.js
@@ -29,13 +29,12 @@ class GuestPortalBinding {
       if (!this.portal) return false
 
       this.sitePositionsController = new SitePositionsController({portal: this.portal, workspace: this.workspace})
-      this.sitePositionsController.show()
+      this.subscriptions.add(this.workspace.onDidChangeActivePaneItem(this.didChangeActivePaneItem.bind(this)))
+      this.subscriptions.add(this.workspace.onDidDestroyPaneItem(this.didDestroyPaneItem.bind(this)))
 
       await this.portal.setDelegate(this)
       await this.toggleEmptyPortalPaneItem()
 
-      this.subscriptions.add(this.workspace.onDidChangeActivePaneItem(this.didChangeActivePaneItem.bind(this)))
-      this.subscriptions.add(this.workspace.onDidDestroyPaneItem(this.didDestroyPaneItem.bind(this)))
       return true
     } catch (error) {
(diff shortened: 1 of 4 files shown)


nathansobo:
If I understand this correctly, it seems a bit counter-intuitive to me. To me, having the avatar follow the cursor around seems to indicate that the user the avatar represents is also following the cursor, which isn't the case. Without having used it yet, I *think* I prefer the previous approach. Maybe we could indicate the location of other users somehow as well to clarify.

I'm wondering if there's some other way of representing the active users and where they are located in the workspace. A "buddy list" dock item could potentially serve that purpose, and by separating it visually from the panes it would prevent the confusion of an avatar being superimposed over non-remote panes. The downside of a dock item is it would consume more screen real estate than the current approach.

Anyway, I think we have more design work to do on this aspect of the UI. @simurai do you have any thoughts on this? How can we indicate the presence of collaborators and provide a convenient affordance for jumping to their location that clearly indicates what is happening and consumes minimal screen real estate.

simurai:
Here some explorations around showing avatars 👉  #286

PR #287, opened 2017-12-15
Provide "reload" button when displaying initialization error

### LINKED ISSUE #250: initialization failed
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:

### Description

[Description of the issue]

### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

**Expected behavior:**

[What did you expect to happen?]

**Actual behavior:**

[What actually happened instead?]

**Reproduces how often:**

[What percentage of the time does this happen?]

### Versions

You can get this information from copy and pasting the output of `atom --version` and `apm --version` from the command line. Also, please include the OS and what version of the OS you're running.

### Additional Information

Please include any additional information, configuration or data that might be necessary to reproduce the issue. Screenshots, if appropriate, are helpful as well.

jasonrudolph:
The initialization errors seen in #250 appear to be transient. https://github.com/atom/teletype/issues/250#issuecomment-348382047 notes:

> On OSX I had `Failed to load resource: net::ERR_NAME_NOT_RESOLVED` but was also unable to check for outdated packages in atom. It may be a DNS race, where it times out before resolving. Loading https://atom.io in a browser and a restart "fixed" it for the time being. Making me think that it's an infrastructure or network error rather than `teletype` specific.

Since "trying again" often Atom seems to resolve the issue, this pull request provides a "Reload Atom" button when the popover component displays an initialization error.

### Demo

demo

### Verification Process

1. Disable network access on computer
2. Open Atom window
3. Verify that "Reload Atom" button appears in Teletype popover
4. Enable network access son computer
5. Click "Reload Atom"
6. Click on the portal icon to verify that Teletype initialized successfully

---

This change may be sufficient to close #250. I propose that we roll this out and see if people still run into issues.

### DIFF
--- a/lib/package-initialization-error-component.js
+++ b/lib/package-initialization-error-component.js
@@ -16,16 +16,16 @@ class PackageInitializationErrorComponent {
   render () {
     return $.div({className: 'PackageInitializationErrorComponent'},
       $.h3(null, 'Teletype initialization failed'),
-      $.p(null, 'Make sure your internet connection is working and reload Atom.'),
+      $.p(null, 'Make sure your internet connection is working and restart the package.'),
       $.div(null,
         $.button(
           {
             ref: 'reloadButton',
             type: 'button',
             className: 'btn btn-primary inline-block-tight',
-            onClick: this.reloadAtom
+            onClick: this.restartTeletype
           },
-          'Reload Atom'
+          'Restart Teletype'
         )
       ),
       $.p(null,
@@ -36,7 +36,9 @@ class PackageInitializationErrorComponent {
     )
   }
 
-  reloadAtom () {
-    location.reload() // eslint-disable-line no-undef
+  async restartTeletype () {
+    const {packageManager} = this.props
+    await packageManager.deactivatePackage('teletype')
+    await packageManager.activatePackage('teletype')
   }
 }
--- a/lib/package-initialization-error-component.js
+++ b/lib/package-initialization-error-component.js
@@ -16,12 +16,27 @@ class PackageInitializationErrorComponent {
   render () {
     return $.div({className: 'PackageInitializationErrorComponent'},
       $.h3(null, 'Teletype initialization failed'),
-      $.p(null, 'Make sure your internet connection is working and restart the package.'),
+      $.p(null, 'Make sure your internet connection is working and reload Atom.'),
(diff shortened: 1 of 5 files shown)


nathansobo:
Clever solution to help mitigate this.

lib/package-initialization-error-component.js
```diff
@@ -16,12 +16,27 @@ class PackageInitializationErrorComponent {
...
       $.p(null,
         'If the problem persists, visit ',
         $.a({href: 'https://github.com/atom/teletype/issues/new', className: 'text-info'}, 'atom/teletype'),
         ' and open an issue.'
       )
     )
   }
+
+  reloadAtom () {
+    location.reload() // eslint-disable-line no-undef
```
jasonrudolph: @as-cii: Is this a reasonable approach? Or is there another implementation that would be more preferable here?

nathansobo: You can also call `atom.reload()`, which RPCs back to the main process. I'm not actually clear on whether there ends up being a material difference. The same exact code may end up running on reload either way. We'd have to investigate to be totally sure.

as-cii:
I wonder if we should simply provide a "Reload Package" facility as opposed to reloading the entire window. When clicked, that button would: 

* Reset the popover state.
* Attempt to re-initialize the package.
* If initialization succeeds, attempt to sign in using the token saved in the keychain.

What do you think? 💭

jasonrudolph:
> I wonder if we should simply provide a "Reload Package" facility as opposed to reloading the entire window.

I like this idea, and I should have listed this as an alternate design. I considered this option before opening this pull request, but I wanted to propose this approach first for a couple reasons:

- It's already implemented, which means that we can focus on higher-impact improvements. 😇 
- It's very little code to maintain. If we instead attempt to implement package-specific reloading logic, I suspect we'll have to touch more areas of the package, and that introduces more cognitive weight each time we want to make a change to those areas of the package.

@as-cii: What do you think?

jasonrudolph:
@as-cii: Following our discussion today, I've updated this pull request to deactivate and reactivate the Teletype package, instead of performing a full Atom reload:

demo

How's this look?

nathansobo:
✨

PR #301, opened 2018-01-09
Update for compatibility with Atom 1.25

### LINKED ISSUE #300: `TypeError: this.historyProvider.groupLastChanges` when joining portal
### Description

Using the latest dev build of Atom, teletype throws an error when a participant attempts to join a portal.

### Steps to Reproduce

1. Host shares a portal
2. Guest joins the portal

**Expected behavior:**

Guest should be in the portal and able to collaboratively edit buffers with the host.

**Actual behavior:**

Guest sees error notification saying:

> Failed to join portal
>
> Attempting to join portal [REDACTED] failed with error: this.historyProvider.groupLastChanges is not a function

**Reproduces how often:**

100%

### Versions

```
$ atom --version
Atom    : 1.25.0-dev-dced608f4
Electron: 1.7.10
Chrome  : 58.0.3029.110
Node    : 7.9.0

$ apm --version
apm  1.19.0
npm  3.10.10
node 6.9.5 x64
atom 1.25.0-dev-dced608f4
python 2.7.10
git 2.15.1

$ apm list | grep teletype
├── teletype@0.5.0
```

nathansobo:
🍐 d with @as-cii 

Closes #300 

Atom 1.25 introduces two big changes that affect the Teletype package.

1. [x] It upgrades to a version of `text-buffer` containing the changes in https://github.com/atom/text-buffer/pull/282, which changes the behavior `undo: "skip"` option when calling `setTextInRange`. The change is now pushed to the history provider to avoid corruption of the default linear history. In addition, we expect the history provider to implement the `groupLastChanges()` method, which groups the last change with the change before it. This PR updates Teletype to no longer assume `undo: "skip"` does not push the change to the history and implements the new method.

2. [x] We upgrade to Electron 1.17. This seems to introduce some changes in the underlying WebRTC implementation that are causing our tests to fail, though the package seems to work in practice. We need to figure this out or our tests will start failing in February.

screen shot 2018-01-09 at 10 18 19 am

/cc @jasonrudolph

### DIFF
--- a/lib/buffer-binding.js
+++ b/lib/buffer-binding.js
@@ -11,6 +11,7 @@ class BufferBinding {
     this.emitDidDispose = didDispose || doNothing
     this.pendingChanges = []
     this.disposed = false
+    this.disableHistory = false
   }
 
   dispose () {
@@ -40,10 +41,15 @@ class BufferBinding {
   }
 
   setText (text) {
+    this.disableHistory = true
+    // TODO: Remove undo skip and require usage latest Atom in engines field when Atom 1.25 reaches stable
     this.buffer.setTextInRange(this.buffer.getRange(), text, {undo: 'skip'})
+    this.disableHistory = false
   }
 
   pushChange (change) {
+    if (this.disableHistory) return
+
     if (this.bufferProxy) {
       const {oldStart, oldEnd, newText} = change
       this.bufferProxy.setTextInRange(oldStart, oldEnd, newText)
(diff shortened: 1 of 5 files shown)


jasonrudolph:
@nathansobo @as-cii: Am I correct in thinking that this PR closes #300?

nathansobo:
Updated. Thanks @jasonrudolph. Was in a hurry to open this before stepping away for a chore and forgot to ref it.

PR #302, opened 2018-01-12
Add Atom dev channel to build matrix

jasonrudolph:
In addition to running CI against the latest stable and beta versions of Atom, also run CI against atom/atom's master branch. This will help us more quickly identify issues like #300, in which a change on atom/atom's master branch breaks Teletype.

/xref atom/ci#80

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,8 +31,7 @@ addons:
 before_script:
   - createdb teletype-test
 
-# TODO: Once https://github.com/atom/ci/pull/80 is merged, update URL below to use build-package.sh on master branch
-script: 'curl -s https://raw.githubusercontent.com/atom/ci/0155aa53b45d00c5b9be9adb33f4a4da1a96ffdb/build-package.sh | sh'
+script: 'curl -s https://raw.githubusercontent.com/atom/ci/master/build-package.sh | sh'
 
 git:
   depth: 10
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ env:
   matrix:
   - ATOM_CHANNEL=stable
   - ATOM_CHANNEL=beta
+  - ATOM_CHANNEL=dev
 
 notifications:
   email:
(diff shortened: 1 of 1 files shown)


jasonrudolph:
To verify that this is working as expected, I pushed up [a temporary branch](https://github.com/atom/teletype/blob/8127134f5f625743d38c09f298dcb4308509b8ef/.travis.yml#L11) that runs CI against the dev channel.

Before #300 was fixed, we see the dev channel build failing with the same errors reported in #300: 
https://travis-ci.org/atom/teletype/jobs/328143208#L524-L526

Now that #300 was fixed in #301, we see the dev channel build passing:
https://travis-ci.org/atom/teletype/jobs/328161857#L524-L526

nathansobo:
Awesome. Does Travis have a way to kick off a nightly build?

nathansobo:
Looks like yes! https://docs.travis-ci.com/user/cron-jobs/

jasonrudolph:
Thanks, @nathansobo. I added a daily cron job:

screen shot 2018-01-12 at 17 39 46 pm

It ran for the first time here: https://travis-ci.org/atom/teletype/builds/328311318

.travis.yml
```diff
@@ -30,8 +31,7 @@ addons:
 before_script:
   - createdb teletype-test
 
-# TODO: Once https://github.com/atom/ci/pull/79 is merged, update URL below to use build-package.sh on master branch
-script: 'curl -s https://raw.githubusercontent.com/atom/ci/a5ca10c32676334605be6492369327056e6692b0/build-package.sh | sh'
+script: 'curl -s https://raw.githubusercontent.com/atom/ci/master/build-package.sh | sh'
```
Arcanemagus: Note that piping a script like this can lead to issues if the connection gets killed in the middle of the script executing and it can be _very_ hard to diagnose as the script output gets quite garbled. That's why `atom/ci` recommends [this way](https://github.com/atom/ci/blob/83744c0308ea7188ff90bd719a86b9dc0fb1d0d0/.travis.yml#L18-L21).

jasonrudolph: @Arcanemagus: Thanks for the tip! https://github.com/atom/teletype/pull/304 adopts `atom/ci`'s recommended approach.

PR #304, opened 2018-01-16
Use atom/ci's recommended commands for fetching & running the build script

jasonrudolph:
In https://github.com/atom/teletype/pull/302#discussion_r161645497, @Arcanemagus suggested an improvement to the `script` section in `.travis.yml`:

> Note that piping a script like this can lead to issues if the connection gets killed in the middle of the script executing and it can be _very_ hard to diagnose as the script output gets quite garbled. That's why `atom/ci` recommends [this way](https://github.com/atom/ci/blob/83744c0308ea7188ff90bd719a86b9dc0fb1d0d0/.travis.yml#L18-L21).

This pull request adopts `atom/ci`'s recommended approach.

---

@Arcanemagus: Thanks for the tip! ⚡️

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,10 @@ addons:
 before_script:
   - createdb teletype-test
 
-script: 'curl -s https://raw.githubusercontent.com/atom/ci/master/build-package.sh | sh'
+script:
+  - curl -s -O https://raw.githubusercontent.com/atom/ci/master/build-package.sh
+  - chmod u+x build-package.sh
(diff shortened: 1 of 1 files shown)


Arcanemagus:
It hasn't been a problem for a while, but for a month or so there Travis (or the version of `curl` on their builds?) was killing connections that didn't have data after a rather short period of time... which would often get triggered while Atom was getting installed so it would read enough of the script to get to the part where it attempted to lint the source code and then blow up in _very_ strange ways 😛.

Issue #238, opened 2017-11-17
Show error message in tooltip when initialization fails


50Wliu:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Followed all applicable steps in the [debugging guide][debugging-guide]
    * Checked the FAQs on the [message board][message-board] for common solutions
    * Checked that your issue [isn't already filed][already-filed]?

[debugging-guide]: http://flight-manual.atom.io/hacking-atom/sections/debugging/
[message-board]: https://discuss.atom.io/c/faq
[already-filed]: https://github.com/atom/teletype/issues

### Description

I don't think there's a reason to create a notification error when initialization fails because you'll always be going through the tooltip UI.  Perhaps the notification was added before the tooltip existed?  Either way, the tooltip should provide a copy of the error message.
![teletype-initialization-failed-no-try-again-button](https://user-images.githubusercontent.com/2766036/32964489-b2d6188a-cbd3-11e7-9fe2-d0dda530732d.png)

### Steps to Reproduce

1. Turn off the interwebs
2. Try to start teletyping

**Expected behavior:** Tooltip explains what the error is

**Actual behavior:** Tooltip contains a generic failure message

**Reproduces how often:** 100%

rvermeiren:
Hi, 
I'm a student in last year in computer science at the UCL in Belgium and I want to make my first contribution to an open source project. I choose atom/teletype as it's one of my favorite tools for collaborative work and pair programming. 

Do you think it's an affordable fix for a beginner? I will work on a VM with the last LTS version of Ubuntu and use the development mode of Atom for testing my fix.

I have already read a lot of documentation on how to contribute to Atom.

I will very thankful if someone could help me if I have a problem.

Regards, 

Rémy

nathansobo:
@rvermeiren Thanks for reaching out. I think it's worth a try. You need to understand how this package updates the contents of the tooltip and also look at where we present the notification. You'll also want to take a look at the existing tests to figure out how to add one to cover the new behavior. Good luck!

jasonrudolph:
Closed by #309. Thanks, @rvermeiren. :bow:

PR #309, opened 2018-01-22
remove error alert when no internet and add error description in the popover

rvermeiren:
Remove error alert when no internet and add error description in the popover

**Please be sure to read the [contributor's guide](https://github.com/atom/teletype/blob/master/CONTRIBUTING.md) before submitting any pull requests.**

### Requirements

### Description of the Change

To avoid to generate an error when the internet is not connected, delete the error throw when initialization fails and move the message in the popover launch when the internet is not connected.

 I also try to change the error thrown when the connection is lost since the initialization but don't have understood yet how it's work.

### Alternate Designs

As suggested in the issue (https://github.com/atom/teletype/issues/238), remove the error thrown when initialization of teletype. And pass the message to the popover. It's suggested because not having internet is not really an error.

### Benefits

An error less

### Possible Drawbacks

Maybe another error can catch and pass to the popover.

### Verification Process

Run atom without internet and with teletype installed. --> Tooltip contains the error and suggest to restart.

Run atom with internet --> correct behaviour.

I can't run the local test on my computer as it seems to be an error with the postgresql database.

### Applicable Issues

**Issue (https://github.com/atom/teletype/issues/238)** : 
Show error message in tooltip when initialization fails

### DIFF
--- a/lib/package-initialization-error-component.js
+++ b/lib/package-initialization-error-component.js
@@ -16,7 +16,7 @@ class PackageInitializationErrorComponent {
   render () {
     return $.div({className: 'PackageInitializationErrorComponent'},
       $.h3(null, 'Teletype initialization failed'),
-      $.p(null, 'Error : ' + this.props.initializationError.message),
+      $.p(null, 'Error: ' + this.props.initializationError.message),
       $.p(null, 'Make sure your internet connection is working and restart the package.'),
       $.div(null,
         $.button(


rvermeiren:
Hi, first, thank you for your attention. This is my first contribution on Github and I'm very thankful for your help and review. 
I have fix the lint issue and the comment code you find.

For the testing, I remove two test because the scenario was wrong. When internet is turned of, initializationError not allow to share portal or join one. So, these scenario were good before because the test catch the old exception in place of the good one (tested in another file). So i just update the good test and remove the two other.

I hope this is ok and my contribution is good. Regards, Rémy

rvermeiren:
Thank you for helping me with my first contribution. I think it was a great experience and I will probably continue to contribute further.

I see that one check not passed with the merge. Can I do something about that ? 

image

50Wliu:
Nice work @rvermeiren!  We would love to see you contribute more :).

You do not have to worry about the pending status check.

PR #311, opened 2018-01-25
Force stats collection to be disabled

damieng:
Stats are currently effectively disabled by the Chrome content security policy but we should [explicitly disable them](https://github.com/pusher/pusher-js/tree/v4.1.0#disablestats-boolean) to get rid of the console log warning and to prevent them accidentally becoming enabled should the CSP or implementation change.

If somebody does want to enable them we would need to:

1. Honor the `telemetryConsent` config setting in Atom - would need to be set to `limited` for opt-in
2. Need to modify the CSP to enable this host to get through Chrome

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,10 +6,10 @@ module.exports = new TeletypePackage({
   commandRegistry: atom.commands,
   tooltipManager: atom.tooltips,
   clipboard: atom.clipboard,
-  disableStats: true,
   pusherKey: atom.config.get('teletype.pusherKey'),
   pusherOptions: {
-    cluster: atom.config.get('teletype.pusherCluster')
+    cluster: atom.config.get('teletype.pusherCluster'),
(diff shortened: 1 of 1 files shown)


jasonrudolph:
@damieng: Thanks for opening a pull request to remove the error that we're seeing in the console. I tried out this change locally, and I'm still seeing the error:

demo

Do the changes in this pull request resolve the error for you?

damieng:
Sorry, had done a git reset to clear out a bunch of changes i'd made to teletype to simulate errors for a different PR/package and when I put the line back in I put it in the wrong place.

jasonrudolph:
Thanks, @damieng. With the changes in 51a5f17941d0e539f6c94c887c8bd0a2a21a3bfa, I no longer see the error in the console when sharing a portal. :bow:

Issue #268, opened 2017-12-11
[Tracking] RFC-001: Allow guests to collaborate on multiple remote buffers


jasonrudolph:
We intend to implement [RFC-001](https://github.com/atom/teletype/blob/master/doc/rfcs/001-allow-guests-to-open-multiple-remote-buffers.md) via a series of pull requests. This will allow developers to start benefiting from the completed RFC-001 functionality without having to wait for _all_ of RFC-001 to be completed. We'll use this issue to track progress.

## TODO

- [x] When following the host, any time they switch to a new buffer, a new editor for that remote buffer is automatically added to your workspace and focused. Existing editors for previous remote buffers are not automatically closed when this switch occurs. (#262) ([v0.3.0](https://github.com/atom/teletype/releases/tag/v0.3.0))
- [x] When a host closes a buffer, it is removed from all guest portals. (#262) ([v0.3.0](https://github.com/atom/teletype/releases/tag/v0.3.0))
- [x] When you are following a guest and the guest navigates to a different buffer in the same portal, you follow them to that buffer. (#262) ([v0.3.0](https://github.com/atom/teletype/releases/tag/v0.3.0))
- [ ] Show site position avatars for participants that are in non-portal pane items (#332)
- [x] Allow guests to use fuzzy-finder to open any remote editor shared by the host (https://github.com/atom/teletype/pull/323)
- [x] Once the fuzzy-finder support is in place, consider removing the logic that causes guests to automatically re-follow the host when the portal goes from *n* to zero remote items. (If the guest is working in a local editor when the portal goes from *n* to zero remote items, and we automatically re-follow the host, it will yank the guest away from their current editor. We'd like to avoid that. 😬) (https://github.com/atom/teletype/pull/323)
- [x] Close remote buffers on guest when portal gets closed or disconnected (https://github.com/atom/teletype/pull/319)
---

Refs: https://github.com/atom/teletype/pull/231 (which introduced RFC-001)

nathansobo:
I'd like to explore our handling of the host closing a buffer. Currently, I think we're planning on warning them if guests are editing. Another option could be to go ahead and remove the editor from the host's workspace, but maintain a reference to the editor in the host's pool (keep it's ref count above zero). This would allow guests to continue to edit and save the buffer.

[...]

PR #319, opened 2018-02-07
Close remote buffers on guest when host closes portal

jasonrudolph:
### Description of the Change

This change addresses the following item from the task list in https://github.com/atom/teletype/issues/268:

> - [x] Close remote buffers on guest when portal gets closed or disconnected (https://github.com/atom/teletype/commit/9d9f74c11a33e81beca11d5cf16ac4725f1e2d5b#commitcomment-26310833)

In the discussion in https://github.com/atom/teletype/commit/9d9f74c11a33e81beca11d5cf16ac4725f1e2d5b#commitcomment-26310833, we observed that it's often "annoying that the host's buffers persist as unsaved files after they close their portal. With one editor, it was fine, but now there are quite a few tabs to clean up after a host terminates their session." With the changes in this pull request, when the host closes the portal, Teletype will clean up the guest's workspace to remove the closed portal's tabs.

### Verification Process

- [x] When the host closes the portal, verify that all the portal's tabs are closed on the guest
- [x] When a guest is participating in multiple portals (A & B), verify that when portal A is closed by the host, portal A's tabs are closed on the guest, but portal B's tabs remain open on the guest
- [x] When the guest leaves a portal, verify that all the portals tabs are closed on the guest (i.e., the same behavior that existed prior to this pull request)
- [x] When the guest closes the last remote editor for a portal, verify that the guest leaves the portal (i.e., the same behavior that existed prior to this pull request)
- [x] When the host closes the last editor in their workspace, verify that the guest sees the empty portal pane item (i.e., the same behavior that existed prior to this pull request)

### DIFF
--- a/lib/editor-binding.js
+++ b/lib/editor-binding.js
@@ -28,14 +28,15 @@ class EditorBinding {
 
     this.markerLayersBySiteId.forEach((l) => l.destroy())
     this.markerLayersBySiteId.clear()
-    if (!this.isHost) {
-      this.restoreOriginalEditorMethods(this.editor)
-      this.editor.destroy()
-    }
     if (this.localCursorLayerDecoration) this.localCursorLayerDecoration.destroy()
 
     this.emitter.emit('did-dispose')
     this.emitter.dispose()
+
+    if (!this.isHost) {
+      this.restoreOriginalEditorMethods(this.editor)
+      this.editor.destroy()
+    }
   }
 
   setEditorProxy (editorProxy) {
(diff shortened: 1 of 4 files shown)


jasonrudolph:
@as-cii: I think this is _almost_ ready to merge, but there's one regression that I'm observing: Prior to this pull request, when the host closes the portal, the guest sees a [notification saying that the portal was closed](https://user-images.githubusercontent.com/2988/35945985-180c8bfc-0c30-11e8-9837-a822509334cb.png). With the changes in this pull request, the guest no longer sees that notification; they just see all the remote tabs get closed. I think it's important for Teletype to show this notification, so that the guest understands why the tabs were closed.

I haven't yet had a chance to look into addressing this regression, but I think that's the next step for moving forward with this pull request.

jasonrudolph:
A quick before-and-after demo of the guest's experience when the host closes the portal:

### Before: So many tabs to close 🙀

before

### After: Such a clean workspace 🧘‍♂️

after

PR #321, opened 2018-02-08
🔥 Remove obsolete logic for "deremotifying" guest editors

jasonrudolph:
Prior to #319, when the host closed the portal, Teletype would transform the remote editors in the guest's workspace into local untitled/unsaved editors. As of #319, when the host closed the portal, Teletype *closes* remote editors in the guest's workspace. As a result, we no longer need the logic for transforming the remote editors into local editors.

### Verification Process

- [x] As a guest viewing multiple remote editors in a portal, verify that all remote editors are closed when the host closes the portal
- [x] As a guest viewing multiple remote editors in a portal, verify that all remote editors are closed when you leave the portal

### DIFF
--- a/lib/editor-binding.js
+++ b/lib/editor-binding.js
@@ -33,10 +33,7 @@ class EditorBinding {
     this.emitter.emit('did-dispose')
     this.emitter.dispose()
 
-    if (!this.isHost) {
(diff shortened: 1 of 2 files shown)


as-cii:
Nice! 🔥

Issue #284, opened 2017-12-14
Feature Request: Remembered, Persistent, and/or Favorited Portals


gwpantazes:
### Prerequisites

* [x ] Put an X between the brackets on this line if you have done all of the following:
    * Followed all applicable steps in the [debugging guide][debugging-guide]
    * Checked the FAQs on the [message board][message-board] for common solutions
    * Checked that your issue [isn't already filed][already-filed]?
        * I've looked and what I'm suggesting is similar and related to these issues: #22, #213, #214, because these issues have to do with remembering a past portal. 

### Description

Can the Teletype plugin keep a portal and remember it so that the next time the user wants to link the same workspace, no UUID has to be entered?

If the plugin could remember a specific portal for the user and the portal could act like a remembered, persistent, and/or favorited portal, this could save time for a user.

My use case that I can bring to the table is that I'd like to share a portal workspace between 2 (or 3!) of my own machines. Since they are my own machines and I routinely plan on connecting them via teletype each morning and throughout the day, a mechanism to make that process faster and smoother (such as the remembered portal I'm suggesting) would be nicer than having to reshare different UUID's each time.

**Expected behavior:**

I would want to see a one time first setup step with some kind of toggle where after setting up the portal with the UUID, each successive portal connection does not need a UUID again, since the portal would be remembered from last time.

**Actual behavior:**

Each portal connection needs a UUID and a brand new portal connection.

### Versions

MacOS Sierra 10.12.6
```
$ atom --version
Atom    : 1.23.1
Electron: 1.6.15
Chrome  : 56.0.2924.87
Node    : 7.4.0
$ apm --version
apm  1.18.11
npm  3.10.10
node 6.9.5 x64
atom 1.23.1
python 2.7.10
git 2.10.1
```

jasonrudolph:
:wave: Hi @gwpantazes! Thanks for taking the time to write this up, and I especially want to thank you for sharing your use case. It's always helpful for us to understand the benefit that people are seeking when discussing a feature request. :bow:

And thank you for cross-referencing #213. It is indeed similar, but your use case is different enough that I think it makes sense to for #213 and this issue to both remain open.

[...]

Issue #213, opened 2017-11-16
Favourites/recent collaborators list


VariableClass:
### Favourites/recent collaborators list 

It would be nice to join or share a portal through some sort of mechanism using GitHub users. 🙂 

For example, selecting the user profile picture of a recently collaborated with/favourite user from within the Teletype interface would send them an e-mail via GitHub and/or flag a notification on the Teletype icon in Atom inviting them to collaborate.

[Atom Discussion Thread](https://discuss.atom.io/t/teletype-favourites-recent-collaborators-list/49752)

jasonrudolph:
Thanks for taking the time to share this idea! :bow:

The concept of a "buddy list" (or something similar) is definitely something that's been on our minds (e.g., #22).

As we prioritize future work, we'll be looking at feature requests like this one to determine where we can deliver the most benefit with the available engineering resources (i.e., full-time Atom engineers, community maintainers, and new contributors). In the meantime, if this is a feature that would make you stand up and dance each time you use Teletype 🕺:smile:, then [adding a :+1: emoji reaction](https://github.com/blog/2119) to the [issue body](https://github.com/atom/teletype/issues/213#issue-274502973) is a great way to let us know.

VariableClass:
Hi @jasonrudolph, I'd say that both this and #214 are covered by #22! I did look through the ___open___ issues to check there was nothing similar already in existence, but not the __closed__ issues. It therefore seemed appropriate that these were two separate feature requests, but I'm happy for these to be closed off if there's some way I can show some ❤️ for #22? 🙂

jasonrudolph:
Hi @VariableClass: Thanks for the follow-up. :bow: I think I might see why you're looking at #213 and #214 as two separate feature requests: Since a portal is permanently closed when the host stops sharing the portal, it seems like #214 might be a prerequisite to #213. First, you need to be able to share/join a portal by specifying a GitHub @username instead of a portal ID (i.e., #214). Once that functionality is in place, we can extend it to let you keep track of the GitHub users that you've recently collaborated with and/or a list of your favorite GitHub users to collaborate with (i.e., #213)?

How does that compare to what you had in mind?

VariableClass:
That sounds more a bit more cohesive, yes 🙂

Issue #109, opened 2017-10-17
Make the portal ID a link that opens in Atom


johndbritton:
I think it’d be worth creating and implementing a file opening handler for an atom portal scheme.

This way when you share a portal ID it would be clickable and it would open with atom automatically.

`atom-portal://portal-is-here`

---

**UPDATE 2018-03-19**: We'll likely tackle this via a series of pull requests, with each one addressing one or more of the items below:

- [x] Teach Teletype to join portal via URL (https://github.com/atom/teletype/pull/346)
    - Add URI handler
    - Automatically join the portal when URL is followed
    - Add command for getting the URL
    - Don't provide any UI for getting the URL. (We'll handle that below.)
- [ ] Ask the user to confirm before joining the portal [[motivation](https://github.com/atom/teletype/blob/e35483b1fa69c3ef1ef56f0c437000d4dcda934b/doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md#join-a-portal-via-url)]
    - Provide option to join
    - Provide option to decline/ignore
    - Provide option to automatically join when following future portal URLs
    - Provide configuration setting to enable/disable this option
- [x] Update Teletype popover UI to provide portal URL to host (#348)
- [x] Update Teletype popover UI to allow guest to enter portal URL to join portal [[motivation](https://github.com/atom/teletype/pull/344#discussion_r175438235)] (#348)
- [x] Gracefully handle edge cases (#348)
    - [x] If user enters a raw portal ID or a malformed URL in the "join portal" dialog, look for a GUID and try to use it as the portal ID
    - [x] If a malformed URL is passed to Teletype's URI handler (e.g., `atom://teletype/whoops`), show error message (e.g., "No portal exists with that URL")
    - [x] Update "No portal exists with that ID" message to speak in terms of URLs instead of IDs

Arcanemagus:
In addition to the references in #33, note that Atom core has URI support now: https://github.com/atom/atom/pull/15935.

BinaryMuse:
For infos:

Relevant PR: https://github.com/atom/atom/pull/11399
Docs: http://flight-manual.atom.io/hacking-atom/sections/handling-uris/

nathansobo:
Looks like this will be doable once 1.23 hits stable thanks to @BinaryMuse and I think it should be high on our list.

jasonrudolph:
We're digging into this starting this week. I've updated the issue body to serve as a tracking issue as we implement the various aspects of this functionality.

[...]

Issue #22, opened 2017-07-13
Buddy list


simurai:
This issue continues from https://github.com/atom/real-time/issues/18#issuecomment-314720119 and is for brainstorming how the buddy list could look and work.

PR #344, opened 2018-03-15
Add RFC for streamlining collaboration set-up

jasonrudolph:
This pull request will deliver an RFC that proposes improvements to streamline collaboration set-up (as [described in our roadmap](https://github.com/atom/atom/blob/a59bb96715f5c26b58cb106acccb4aed2916a6ee/docs/focus/README.md#2-streamline-collaboration-set-up)).

👉👉👉 [Rendered RFC](https://github.com/atom/teletype/blob/add-rfc-for-streamlining-collaboration-setup/doc/rfcs/004-quickly-collaborate-with-coworkers-and-friends.md) 👈👈👈

### TODO

- [x] Complete initial draft
- [x] Post reply requesting comments -- https://github.com/atom/teletype/pull/344#issuecomment-374632372
- [x] Request feedback from GitHub's Community and Safety team -- https://github.com/atom/teletype/pull/344#issuecomment-374638633
- [x] Incorporate feedback from initial reviews
- [x] Request follow-up review from GitHub's Community and Safety team -- https://github.com/atom/teletype/pull/344#issuecomment-377577746
- [x] Incorporate feedback from follow-up reviews
- [x] Resolve all TODOs that are present in the diff
- [ ] Propose order in which we want to tackle the various aspects (either in the RFC document or in a separate tracking issue) -- https://github.com/atom/teletype/pull/344#pullrequestreview-104579154

---

Refs: #109 - Make the portal ID a link that opens in Atom
Refs: #213 - Feature Request: Favourites/recent collaborators list
Refs: #284 - Feature Request: Remembered, Persistent, and/or Favorited Portals
Refs: #22 - Buddy list


jasonrudolph:
@simurai: Would you be willing to start exploring designs and sharing mockups for the functionality described in this RFC? I think that will help us identify the places where the RFC is overly vague and it will also be a big help in communicating these ideas when we open this up for feedback from a wider audience. (Please feel free to challenge any concepts that you see proposed here. I always appreciate your perspective.)

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,118 @@
...
+
+When your past collaborators have Atom open and they're signed into Teletype, they appear as "online" in your list of past collaborators. You can select a past collaborator (an "invitee" in this context) and invite them to join your portal. If you're not already hosting a portal, Teletype automatically creates a portal for your workspace. (Note: We probably need to draft a proposal for implementing "presence." Maybe that belongs in this RFC, or maybe it belongs in a separate one.)
+
+Inside the invitee's Atom instance, Teletype informs the invitee that you have invited them to join your portal, and Teletype asks them if they want to join it. If the invitee chooses to join the portal, they see a notification confirming that they've joined the portal. If the invitee declines to join the portal, Teletype notifies the host that the guest declined the invitation.
+
+### Preventing invitations from past collaborators
+
+You can remove anyone from your list of past collaborators. Once you've removed someone from your list of past collaborators, you will appear as "offline" in their list of past collaborators, and they can no longer invite you to their portal from directly within Atom.
```
simurai: Can the host remove someone from their list while collaborating is still ongoing? I think that would be handy. Like if you invite a bunch of people and then it turns out somebody behaves badly. To that person it could just appear like the host went offline/closed the portal without some drama inducing "you got kicked out" message.

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,118 @@
...
+- What unresolved questions do you expect to resolve through the implementation of this feature before it is released in a new version of the package?
+    - *TODO*
+- What related issues do you consider out of scope for this RFC that could be addressed in the future independently of the solution that comes out of this RFC?
+    - *TODO*
+
+## Even more TODOs 😇
+
+- Explicitly state that past collaborators have two possible states: online or offline. Setting your status to "away" or "busy" is out of scope for now. Making yourself invisible is out of scope for now. In the meantime, you sign out of Teletype or disable Teletype in Atom's package settings when you want to use Atom while not accepting invitations from your past collaborators.
+- What happens if I'm currently a guest and I click to invite someone to a portal?
+- How should we handle an extremely large list of past collaborators? How should we sort it? Should the user be able to change how it's sorted? Should they be able to filter the list somehow?
```
simurai: - Sort by how recently you collaborated with someone.
- Automatically remove people if you haven't collaborated for a long time. Like 2-3 months. This might feel intrusive, but at the same time it let's you keep inviting more people without having to worry about cleaning up your list. Your list stays current automatically. And you can always invite people again.
- Then the list stops at like 50 people? With a filter input at the top to quickly find someone.
- The filter input only shows if you have more than 10 people in your list, otherwise it feels unnecessary.

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Streamline collaboration while avoiding harassment
+
+With the approach described above, we believe Teletype can provide a streamlined set-up process while also preventing harassment. In order for Teletype to prompt you to join someone's portal, you must first have collaborated with that person in the past (by sending your portal URL to them via a third-party service or by joining their portal via a URL they sent you via a third-party service) *and* that person must still exist in your list of past collaborators (i.e., you haven't removed them from your list of past collaborators). In other words, people that you've never collaborated with cannot cause portal invitations to appear in your editor. And people that you've collaborated with in the past, but don't want to receive invitations from in the future, cannot cause portal invitations to appear in your editor either.
+
+## Drawbacks
+
+Why should we *not* do this?
+
+*TODO*
```
simurai: - Privacy? Some people might feel uneasy knowing that other collaborators can see if they are "online" a.k.a. working in Atom. As mentioned in the "Even more TODOs", there could be an offline mode.

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
jasonrudolph: Thanks, @simurai. I :heart: these ideas. 38a3b71875d1ab4e14ef1c1852cabbead343791f proposes an MVP for the initial implementation, but your notes above feel like a solid direction to move in once we have the barebones functionality in place.

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,126 @@
...
+1. Host shares a portal
+2. Host selects person in their list of past collaborators
+3. Host invites person to join the portal
+4. Guest accepts invitation to join portal, and Atom joins the portal
+
+And perhaps we can reduce the host's set-up process even further:
+
+1. Host selects a person in their list of past collaborators
+2. Host invites the selected person to a portal, and Atom automatically shares a portal for the host and invites the selected person
+3. Guest accepts invitation to join portal, and Atom joins the portal
```
as-cii: I really like how you structured this paragraph. I am pretty excited to see how we can reduce the current 8 steps to just 3 or 4! 💯

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Explanation
+
+### Portal URLs
+
+*TODO* Define URL format
```
as-cii: I know this still needs to be written, but it might be worth mentioning that we currently use `atom://teletype/portal/123/editor/345` to represent editor URIs in Atom's workspace.

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Share a portal via URL
+
+Once a host creates a portal (e.g., by clicking the "share" toggle in Teletype's status bar popover), Teletype presents the host with the portal's URL. The host copies that URL to their clipboard (in the same way that they previously copied the portal ID).
+
+#### Early mockup
+
+<img width="314" src="https://user-images.githubusercontent.com/2988/37489877-6d0b0cfe-286f-11e8-87a3-a04da42994a3.png">
+
+Also consider [Slack's approach](https://github.com/atom/teletype/issues/109#issuecomment-337223084) for potential inspiration.
```
as-cii: I really like Slack's design solution to this. 👍

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Streamline collaboration while avoiding harassment
+
+With the approach described above, we believe Teletype can provide a streamlined set-up process while also preventing harassment. In order for Teletype to prompt you to join someone's portal, you must first have collaborated with that person in the past (by sending your portal URL to them via a third-party service or by joining their portal via a URL they sent you via a third-party service) *and* that person must still exist in your list of past collaborators (i.e., you haven't removed them from your list of past collaborators). In other words, people that you've never collaborated with cannot cause portal invitations to appear in your editor. And people that you've collaborated with in the past, but don't want to receive invitations from in the future, cannot cause portal invitations to appear in your editor either.
```
as-cii: This feels like a really good solution. 👍

jasonrudolph:
@nathansobo @as-cii @simurai: I've completed the first draft of this RFC. Before requesting comments from a broader audience, I'd love to get your review:

- What would you change about this RFC in its current form?
- What additional questions/topics should this RFC address?
- Where do you think should we address the architectural questions related to presence? (Maybe we address that via one or more spikes?)

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,146 @@
...
+
+Portal URLs use that same pattern, but exclude the editor ID: `atom://teletype/portal/<portal-id>`
+
+The `<portal-id>` is the same ID used in prior releases for sharing and joining a portal.
+
+Example: `atom://teletype/portal/63b120f3-b646-4c46-8962-656518249186`
+
+### Join a portal via URL
+
+When a guest follows the URL (e.g., by clicking on the URL in Slack, IRC, etc.), Atom opens, and Teletype asks the user if they want to join the portal. If the user chooses to join the portal, they become a guest in the portal (just as they previously did by entering the portal ID and clicking "Join"). If they choose not to join the portal, nothing happens.
```
simurai: Should we keep the "Join a portal" button (and input when clicked) in the popup? With clickable URLs it's not really needed, except maybe if the URL handling doesn't work for some reason. Like a mis-configured URL handler. Then there is always a fallback of manually pasting the URL. Is that a big enough reason?

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,146 @@
...
+
+To honor the [UX guidelines for Atom URI handlers](https://flight-manual.atom.io/hacking-atom/sections/handling-uris/), Teletype avoids automatically joining the portal. When asking the user whether they want to join the portal, Teletype offers an option to automatically join future portals. This option is disabled by default. When the user enables this option, any time they follow a portal URL, Teletype will automatically join the portal without the user having to perform an additional confirmation of their desire to join the portal. Users can disable this option at any time via the Teletype settings in Atom's Settings UI.
```
simurai: When a guest joins a new portal by clicking on the "Join portal" button in the Atom notification, does Teletype open a new window or use the same window?

Now that multiple files are supported, I might not want them to get mixed with all the open tabs and fuzzy finder of my current workspace and rather have an empty window during this collaboration.

The notification could also have two options (buttons) to join a portal?

1. Join portal in this window
2. Join portal in a new window

Probably better would be a ✅ "Open in new window" checkbox that can be toggled, but I don't think the Atom notifications API supports having checkboxes?

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,146 @@
...
+- Depending on the technical solution we choose for implementing presence, we may incur increased server-side resource consumption and/or we may take on additional operational complexity.
+
+## Rationale and alternatives
+
+##### Why is this approach the best in the space of possible approaches?
+
+With the approach described above, we believe Teletype can provide a streamlined set-up process while also preventing harassment. In order for Teletype to prompt you to join someone's portal, you must first have collaborated with that person in the past (by sending your portal URL to them via a third-party service or by joining their portal via a URL they sent you via a third-party service) *and* that person must still exist in your list of past collaborators (i.e., you haven't removed them from your list of past collaborators). In other words, people that you've never collaborated with cannot cause portal invitations to appear in your editor. And people that you've collaborated with in the past, but don't want to receive invitations from in the future, cannot cause portal invitations to appear in your editor either.
```
simurai: For a while I was questioning: "Why is the list of past collaborators even needed, when there is no audio support or some sort of text based chat inside Atom?" Because if I invite somebody through the past collaborators list, after they join, I still need to go to Slack, etc. to chat or start an audio call. And if I have to do that, then I could just invite people by pasting the URL there. So it would make more sense to have a roadmap like this: 1. Add invite with URLs 2. Add audio/chat 3. Add past collaborators list.

But here some use cases where a past collaborators list would be handy even without audio/chat:

- When collaborators are in the same room. A host can just invite them directly.
- In recurring 1:1 meetings or standups where people join a video chat anyways. Then a host doesn't have to do a detour and be like "hey, here is today's portal" and can just invite people directly.
- Some people might still use traditional phones or things like FaceTime where you can't paste a link.
- You might start talking to people without intending to use Teletype. But then when you do, both, host and guest can go straight into Atom and the guest doesn't have to wait for the URL to be pasted into a chat.
- Not sure how often people do it, but you kinda can "chat" in an editor by just writing code comments to each other. I think when Teletype launched, and somebody made an ID public, people were "chatting" just to test Teletype.

Ok, so yeah.. maybe still enough good reasons to already have a past collaborators list now.

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,146 @@
...
+
+People will collaborate less often. Given the additional steps needed to start collaborating, there will be more instances where people decide that it's not worth the effort (i.e., people will enjoy rich collaboration less often).
+
+## Unresolved questions
+
+##### What unresolved questions do you expect to resolve through the RFC process before this gets merged?
+
+- If I invite a past collaborator to join my portal, and that person has multiple Atom windows open, can we show the invitation only in the frontmost window?
+- If I invite a past collaborator to join my portal, and that person has Atom windows open on multiple computers, are we OK with showing the invitation on each computer?
+- Where will the list of past collaborators appear in the UI? How do I show/hide this list?
```
simurai: Could the list still be part of the "popup" (click on the Teletype icon in the status bar)? The list of past collaborators and the list of current collaborators could be the same thing. Just that current collaborators would move to the top and instead of a "Invite" button, they would have a "Follow cursor" button.

An alternative could be a separate panel/dock item, but that feels a bit overkill with the current features. Having the list in the popup is nice because it auto-closes when done inviting and doesn't compete with the tree-view or so. I'll make a mockup so we can see how that feels.

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,146 @@
...
+    - Setting your status to offline. (In the meantime, you can sign out of Teletype or disable Teletype in Atom's package settings when you want to use Atom while not accepting invitations from your past collaborators.)
+    - Setting your status to "away" or "busy"
+- UX enhancements to the list of past collaborators
+    - Changing the sort order for your list of past collaborators (e.g., sort by how recently you've collaborated, sort or group by online/offline status)
+    - Limiting the size of your list of past collaborators. (In the meantime, you can remove past collaborators to reduce the size of the list.)
+    - Filtering your list of past collaborators
+    - Marking certain collaborators as "favorites"
+- Offering to install Teletype if you click on a portal URL and don't yet have Teletype installed ([#220](https://github.com/atom/teletype/issues/220))
+- Selecting a past collaborator and asking to join their portal
+- Selecting multiple past collaborators and inviting them to your portal simultaneously. (In the meantime, you can select a collaborator and invite them, then select another collaborator and invite them, and so on.)
```
simurai: That would be nice for use cases like recurring stand-ups or meetings. But it's probably easiest to just share the URL when there are many people.

Here another idea:

Past collaborators shown as group. It's pretty much the same as Slack's DMs list where it can be just a single person or a group (Don and Amanda):

image

That way a host can easily continue a collaboration with multiple people. Like "let's continue tomorrow at 10am". Then the next day, the host can invite the whole group at once. Although this only makes sense if the list is sorted by most recent. Otherwise you wouldn't know which person gets used for an alphabetic list.

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
simurai: Mockup of a combined current and past collaborator list:



- After sharing your portal, you see the list of past contributors.
- Click on "Invite" to invite someone.
- The "Invite" button turns then into a "Invitation sent" status and that collaborator moves to the top.
- After a collaborator joins, the "Invitation sent" status turns into a "Follow" button. It's the same as clicking an avatar in the editor to follow someone's cursor.
- If the invitee doesn't accept the invitation, the host see a "Invitation rejected" status.
- Alternatively or in addition to clicking the "Invite" buttons, a host can also copy the URL at the bottom and share anywhere.
- When the host clicks on the "Share" toggle, the portal is closed again.

And a click through:

portal

There are still a few open questions, like should the list already be visible, even without sharing? Or can you re-invite someone when they don't join or reject. But we can get to those if we like this direction.

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
jasonrudolph: @simurai: I've been thinking about that, too. Some users may choose _not_ to have Atom handle `atom://...` URLs. For that reason, I do think we should keep the "Join a portal" button and input in the popup. And I think you should be able to paste the URL into that input field in order to join the portal. However, when you click the "Join a portal" button in the popup, we may want to include a message explaining the easier option (i.e., let Atom handle the URL for you).

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
jasonrudolph: Great points, @simurai. Thanks for this thoughtful comment. 

e35483b1fa69c3ef1ef56f0c437000d4dcda934b adds a blurb regarding this topic.

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
jasonrudolph: > Here another idea: Past collaborators shown as group. 

I dig that as a potential future enhancement.

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
jasonrudolph: @simurai: Thanks for this mockup. I really dig where this is going. :zap::bow:

> There are still a few open questions...But we can get to those if we like this direction.

I like this direction. 😄 Would you mind taking a stab at those open questions?

A few questions for you about the current mockup:

1. I think we'll still want to provide a way to join a portal via the UI (https://github.com/atom/teletype/pull/344#discussion_r175438235). What would you think about keeping the existing "Join a Portal" option at the bottom of the popup?

    

2. Can you share a mockup for removing someone from your list of past collabortors?

3. It's rare (but possible) to host a portal while also being a guest of a portal:

    

    Can you share a mockup that shows this scenario?

4. Assuming that we want to get the initial release in people's hands [without making people wait for us to implement sorting, filtering, favoriting](https://github.com/atom/teletype/blob/e35483b1fa69c3ef1ef56f0c437000d4dcda934b/doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md#out-of-scope), can you offer any thoughts on how the UI should gracefully handle long lists of collaborators in the meantime?

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
jasonrudolph: Thanks for bringing this up. :zap:

> The notification could also have two options (buttons) to join a portal?

I can see the benefit of that. If we can find a way to present the options nicely in the UI, then I'm all for it. We could offer 3 buttons, but that feels overcomplicated to me:



Suddenly I'm having flashbacks to every restaurant in California.

easy as 1-2-3

"So many decisions. Was this avocado toast really worth all this effort?" :innocent::laughing:

All joking aside, I'm tempted to ship the first release with just two buttons: "Join" and "Decline". Then, once we've used it for a while, and other folks have had a chance to use it as well, we can assess the value/priority of offering other options. What do you think?

nathansobo: Opening a new window is going to be non-trivial to implement, and the window will also open more slowly than would be ideal due to Atom's load time issues. That said, I do think that a remote workspace really *should* appear in its own window. I think the new window option should be considered out of scope for this RFC.

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
nathansobo: Second the value of this without voice. There are times where I *still* don't join a portal when I should because I don't want to perform the copy-and-paste tea ceremony. Simply clicking a button will make this way more accessible.

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,147 @@
...
+
+##### What other approaches have been considered and what is the rationale for not choosing them?
+
+- Teletype could allow you to enter a person's GitHub username (or their email address) to invite them to your portal. This would remove the need for sharing a URL via a third-party service in order to collaborate, and it would reduce the need to provide a list of past collaborators (i.e., you could just enter a username each time you want to collaborate). However, it would make it possible for any GitHub user to cause invitations to appear inside your Atom instance. This introduces a vector for harassment, so we're avoiding this approach.
```
nathansobo: I think this is a really clever way of structuring this to avoid harassment, and it makes complete sense as a place to start. Eventually, I would *still* really like to offer a solution where I can reach any GitHub user via their user name. It seems like giving someone the option to block a single invite or block all invites from first time collaborators might be enough of a recourse in the case of abuse. Then someone would have receive at most *one* unwanted invitation before disabling it. However, for now I think your solution is elegant and anything further would build upon it.

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,147 @@
...
+    - Some users belong to organizations with thousands of members. Just because you're a member of the same organization as someone else doesn't mean that you're comfortable seeing portal invitations from them.
+- Teletype could add support for audio before adding support for inviting past collaborators to your portal. When collaborating with people in different physical locations, you'll likely need audio support in order to collaborate. Since Teletype doesn't currently provide audio support, these collaborators will still need a third-party solution for audio. If you're already relying on a third-party solution for audio, you can often use that same solution to send your collaborators your portal ID, which reduces the need for Teletype providing a list of past collaborators. However, there are still many [scenarios where it's helpful to quickly invite a past collaborator even if Teletype doesn't yet support audio](https://github.com/atom/teletype/pull/344#discussion_r175319913).
```
nathansobo: A bit off-topic: I actually think that *screen sharing*, where I can follow a collaborator outside of Atom, would be more valuable to me than audio.

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,147 @@
...
+
+##### What is the impact of not doing this?
+
+People will collaborate less often. Given the additional steps needed to start collaborating, there will be more instances where people decide that it's not worth the effort (i.e., people will enjoy rich collaboration less often).
+
+## Unresolved questions
+
+##### What unresolved questions do you expect to resolve through the RFC process before this gets merged?
+
+- If I invite a past collaborator to join my portal, and that person has multiple Atom windows open, can we show the invitation only in the frontmost window?
```
nathansobo: We could potentially show the pending invitation in all of their windows, and honestly we're going to have a hard time doing anything else without coming up with some kind of RPC mechanism between package instances. Slightly better would be to show the invitation in all windows, then dismiss it from all windows once it is accepted in one of them. That could probably happen via the server somehow.

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,147 @@
...
+- If I invite a past collaborator to join my portal, and that person has Atom windows open on multiple computers, are we OK with showing the invitation on each computer?
```
nathansobo: The same approach mentioned in my comment about multiple windows applies here.

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,147 @@
...
+- Where will the list of past collaborators appear in the UI? How do I show/hide this list?
+
+##### What unresolved questions do you expect to resolve through the implementation of this feature before it is released in a new version of the package?
+
+- What architecture/services/libraries will we use to implement presence?
```
nathansobo: I'd really like to be a part of this conversation. Pusher has some offerings, and that might be the most prudent course of action in terms of minimizing operational complexity.

I think we could maybe manage dealing with web sockets ourselves if we off loaded communication between the instances to a third party. Redis queues on a hosted provider like [Redis Cloud](https://elements.heroku.com/addons/rediscloud). Then we can communicate between our servers via Redis and handle web sockets in a fairly stateless way. It might be doable. But as one of three people holding the pager it does give me pause.

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,147 @@
...
+- In order to provide the functionality described above, does teletype-server need to persist your list of past collaborators, or can we meet these needs while only storing this data locally?
```
nathansobo: I think to use the past collaborators as an effective anti-harassment mechanism, we'd probably need to store it server side. Otherwise you could counterfeit an invitation to any user you wanted. Maybe I'm missing something though.

nathansobo:
This seems like a solid experience you're proposing here. I added some comments, but seeing the holes in plans isn't really my strength. As far as I'm concerned this sounds great and I'm looking forward to using it.

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
jasonrudolph: [Agreed.](https://github.com/atom/atom/blob/079a43f7bee69a511955e51e9614c3eb12c01b2e/docs/focus/README.md#looking-farther-ahead-2) :smile:

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
simurai: > Suddenly I'm having flashbacks to every restaurant in California.

:joy:

Not sure if technically possible, but this might work:

1. You get a notification to join a portal.
2. You press `cmd-shift-n` to open a new window.
3. The notification appears in the new window too and you click "join portal".

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
as-cii: Yeah, I guess a clean way of modeling this would be to introduce the notion of an invitation on the server that can be canceled somehow, and send a pub-sub notification to the invitees saying that the invitation has expired. The side benefit of doing this is also that hosts could be allowed to cancel pending invites, in case they've clicked on the wrong person.

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
as-cii: Yeah, and storing things on the server has the benefit of working correctly across different Atom installations. At the same time it will introduce privacy concerns, although I am not sure we should be terribly worried about that if we are transparent about it.

That said, there may be a way of storing past collaborators client-side in a way that prevents abuse. Specifically, the teletype package could reject any invitation from users that are not stored locally. It seems like that should work and can't be counterfeited, because the person that is potentially being harassed is in control of whether the invitation should be accepted or not. Identity is always validated by the server, so the client in question can trust that the "inviter" identity is authentic.

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,147 @@
...
+- If the host invites a past collaborator to their portal and then closes the portal before the invitee joins the portal, what should an invitee see/experience to inform them that the portal no longer exists?
```
as-cii: I think modeling this invitation interaction on the server could once again solve this. Upon disconnection, the server could query all the invitations from the disconnecting user and mark their invitations as expired.

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
as-cii: Agreed. This needs a more in-depth conversation.

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
simurai: 1. Ok, yeah, let's keep the "Join a portal" button
2. There could be an X on right of each past collaborator, but that feels a bit too aggressive and can easily be clicked by accident. So how about have a right-click context-menu? Then we can also explain the action in words. -> "Remove collaborator". Sorry, @as-cii :stuck_out_tongue_winking_eye: 
  image
3. Oh.. right. And it seems as a guest, you can even join multiple times. You have to be a very good multi-tasker, though. Ok, to not take up much more space, we could keep joined guests as is -> Only show the avatar next to the host. I think that's fine, because you probably know who the guests are since you just invited them. So it could look like this:
  host guest 2
4. The list of past collaborators has a `max-height`, if the list is longer, you can scroll. We can play with the number, but maybe so that 5 collaborators fit? The scrolling area includes sent and rejected invitations. Then maybe stop rendering at like 100 collaborators, just as a safeguard. If your list is larger than that, you would need some sort of search/filter anyways.
  host

simurai: Now about those other open questions:

- **Should the list already be visible, even without sharing?** I guess? If you click on "Invite" before starting to share, it would start share automatically.
  image
- **Can you invite someone if they are offline?** Yes, absolutely. Having to check back every few minutes to see if they are online would be a pain. As soon as the invitee opens Atom, the notification appears. Seeing that someone is offline is more FYI that it might take a while till that collaborator joins.
- **Can you re-invite someone when they don't join or reject?** No. There must be reasons why someone doesn't join, no need to bug them even more. If you REALLY have to, you could stop and start sharing, and then you can invite again. Or probably better, send them a chat message or so.

jasonrudolph: Thank you, @simurai! 💟

I've incorporated these mockups and scenarios into the RFC.

jasonrudolph:
> - [x] Post reply requesting comments

📣 This RFC is now ready for comments from any interested parties. You can see a rendered version of the current draft here:

https://github.com/atom/teletype/blob/6b70beecae/doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md#quickly-collaborate-with-coworkers-and-friends

I look forward to your feedback.

jasonrudolph:
> - [ ] Request feedback from GitHub's Community and Safety team

:wave: Hi @dmleong, Hi @benbalter: This RFC proposes a mechanism for a person to invite other people to collaborate with them in Teletype, and we want to ensure we're taking reasonable measures to avoid opportunities for abuse and harassment. Can you or a member of your team please review this proposal and share any feedback you might have from a community and safety perspective? If there's any additional information or context that would be helpful, please let me know. :bow:

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
dmleong: Based on Community & Safety's belief in [consensual software](https://medium.com/consensual-software/consensual-software-how-to-prioritize-user-safety-369f5a4dee8), I would strongly suggest keeping the "Join a portal" button to ensure the user's explicit consent, particularly since there are some privacy concerns mentioned below

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
dmleong: > When the user enables this option, any time they follow a portal URL, Teletype will automatically join the portal without the user having to perform an additional confirmation of their desire to join the portal.

Is this going to be specific per user or is this going to be a global configuration? My preference would be per user (i.e. "Would you like to automatically join any sessions with `@UserA`?") as opposed to a global configuration (i.e. "Automatically join all Teletype pairing sessions for all time")

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,172 @@
...
+
+To honor the [UX guidelines for Atom URI handlers](https://flight-manual.atom.io/hacking-atom/sections/handling-uris/), Teletype avoids automatically joining the portal. When asking the user whether they want to join the portal, Teletype offers an option to automatically join future portals. This option is disabled by default. When the user enables this option, any time they follow a portal URL, Teletype will automatically join the portal without the user having to perform an additional confirmation of their desire to join the portal. Users can disable this option at any time via the Teletype settings in Atom's Settings UI.
+
+### Invite a past collaborator to your portal
+
+Teletype provides a list of your past collaborators (i.e., a "buddy list" of sorts). Each time someone joins your portal, Teletype adds them to your list of past collaborators. Each time you join a portal, Teletype adds the host (and any other guests) to your list of past collaborators.
```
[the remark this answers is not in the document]
dmleong: ~~Are there going to be ways to remove past collaborators?~~ (answered below) Block them? 

My thinking is this is ripe for abuse, in addition to the reasons @benbalter listed above: 

1) `@UserA` pairs with me once and it's a great experience!
2) Something happens and we are no longer friends 
3) `@UserA` can automatically open a portal to me any time they want, change their username to a harassing name, or change their avatar to something unsavory

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,172 @@
...
+
+Teletype presents your list of past collaborators sorted alphabetically by username.
+
+When your past collaborators have Atom open and they're signed into Teletype, they appear as "online" in your list of past collaborators.
```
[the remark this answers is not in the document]
dmleong: I would strongly suggest that we do not add this feature to Atom. It's a large abuse vector to have active state shown to others. It allows harassers to know exactly when a person is online and available to be harassed.

i.e. if I have a stalker and I've accidentally clicked on a link they sent to me, they now know my active state. They can know if I'm at work or near a computer. They can then use that information to map my habits or send me messages and then ask why I'm ignoring them when I'm clearly online. 

Again, the ability to block and remove collaborators is very important here!

Additionally, if we are relying heavily on 3rd party apps for link sharing, is this feature needed?

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,172 @@
...
+
+You can select a past collaborator (an "invitee" in this context) and invite them to join your portal. If you're not already hosting a portal, Teletype automatically creates a portal for your workspace.
+
+Inside the invitee's Atom instance, Teletype informs the invitee that you have invited them to join your portal, and Teletype asks them if they want to join it. If the invitee is offline when you invite them to join your portal, they'll see your invitation the next time they open Atom.
+
+If the invitee chooses to join the portal, they see a notification confirming that they've joined the portal. If the invitee declines to join the portal, Teletype notifies the host that the guest declined the invitation.
```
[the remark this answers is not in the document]
dmleong: In addition to the fine points @benbalter made above, might I also suggest we don't show that a guest has declined the invitation. 

Harassers often get instant gratification when a negative interaction happens. (i.e. some folks on Twitter take it as a badge of honor when their accounts get suspended) By removing the feedback loop (the declined invitation) we lower the incentive to harass another person. I suggest we keep this as a "pending invitation" or delete the invitation if the inviter has been blocked

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,172 @@
...
+
+If the invitee declines your invitation, you cannot re-invite to the same portal. There's probably a reason why they didn't join, so there's no need to bug them even more. If you think they declined by accident, reach out to them via a third-party chat service and send them the URL to join your portal.
```
dmleong: > There's probably a reason why they didn't join, so there's no need to bug them even more. If you think they declined by accident, reach out to them via a third-party chat service and send them the URL to join your portal.

I suggest we don't include this in the documentation so as not to encourage this behavior. I :heart: that you cannot re-invite a declined person!

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,172 @@
...
+
+### Prevent invitations from a past collaborator
+
+You can remove anyone from your list of past collaborators. To do so, right-click on them and remove them:
+
+![Mockup: Remove Past Collaborator](https://user-images.githubusercontent.com/378023/37639861-67778f8e-2c56-11e8-994f-428802fa9191.png)
+
+Once you've removed someone from your list of past collaborators, you will appear as "offline" in their list of past collaborators, and they can no longer invite you to their portal from directly within Atom.
+
+If you later decide that you want to collaborate with them again, you can send them a URL (via Slack, IRC, etc.) to join your portal, or you can ask them to send you a URL (via Slack, IRC, etc.) so that you can join their portal. Once you collaborate in a portal together again, Teletype adds them back to your list of past collaborators, and they can once again see your online/offline status in their list of past collaborators.
```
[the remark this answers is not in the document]
dmleong: Again, I think this is a good spirit, but it may not be needed in the documentation (or worded carefully) to lower chances of potential abuse

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,172 @@
...
+While probably not very common, it's possible to host a portal while also participating as a guest in other portals.
+
+![](https://user-images.githubusercontent.com/378023/37640529-6a5ddaac-2c59-11e8-82f5-9ee4c2ed42a5.png)
+
+## Out of scope
+
+In the interest of getting the highest impact functionality in users' hands as quickly as possible and then iterating based on real-world feedback, the following functionality is out of scope for this RFC, but may be addressed in follow-up releases.
+
+- Changing your online/offline status
+    - Setting your status to offline. (In the meantime, you can sign out of Teletype or disable Teletype in Atom's package settings when you want to use Atom while not accepting invitations from your past collaborators.)
```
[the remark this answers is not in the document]
dmleong: > In the meantime, you can sign out of Teletype or disable Teletype in Atom's package settings when you want to use Atom while not accepting invitations from your past collaborators

Is this going to be explicitly clear when you are broadcasting active state? I would not want to broadcast my Teletype state by accident and feel it is a breach of privacy when taken by surprise. I would suggest making it very clear how to disable this feature

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,172 @@
...
+    - Setting your status to "away" or "busy"
+- UX enhancements to the list of past collaborators
+    - Changing the sort order for your list of past collaborators (e.g., sort by how recently you've collaborated, sort or group by online/offline status)
+    - Limiting the size of your list of past collaborators. (In the meantime, you can remove past collaborators to reduce the size of the list.)
+    - Filtering your list of past collaborators
+    - Marking certain collaborators as "favorites"
+- Offering to install Teletype if you click on a portal URL and don't yet have Teletype installed ([#220](https://github.com/atom/teletype/issues/220))
+- Selecting a past collaborator and asking to join their portal
```
dmleong: I am concerned about this potential feature. If they have removed you as a past collaborator, then is this feature overriding their preference?

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,172 @@
...
+- Selecting multiple past collaborators and inviting them to your portal simultaneously. (In the meantime, you can select a collaborator and invite them, then select another collaborator and invite them, and so on.)
+- Providing option to join portal in a new window [[discussion](https://github.com/atom/teletype/pull/344#discussion_r175569812)]
+
+## Drawbacks
+
+- Some people might feel uneasy knowing that past collaborators can see if they are "online" (i.e., if they have Atom open and they're signed into Teletype).
+- Once Teletype provides a list of past collaborators, people may want to be able to chat with those collaborators from within Teletype. While we may eventually want to support chat, any chat-related functionality must exist in service of Teletype's primary vision of "making it as easy to code together as it is to code alone." We'll need to be diligent to avoid scope creep.
```
dmleong: Chat functionality is a huge concern, especially if we do not have a blocking mechanism in place and we are broadcasting active state! 🙅  I'm more than happy to discuss concerns whenever we get to this point 💭

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
dmleong: It's possible to do this, with some caveats: 

- Teletype needs to respect blocks, spammy status, and suspended status 
- Mutual follows could be a strong indicator of trust
- we should still drop them into the "Join pairing session" flow and ensure explicit consent every time

Getting even one unwanted invitation can be traumatic, so I would prefer we reduce this abuse vector as much as possible 😉

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,172 @@
...
+- Teletype could allow you to invite other users that you're already associated with in some way (e.g., fellow collaborators on a GitHub repository, fellow members of a GitHub organization or team). This would remove the need for sharing a URL via a third-party service in order to collaborate, and it would reduce the need to provide a list of past collaborators, and it would reduce the harassment vector described in the previous bullet. However, it would introduce tradeoffs that we'd prefer to avoid:
+    - Teletype would need additional permissions (i.e., [OAuth scopes](https://developer.github.com/apps/building-oauth-apps/scopes-for-oauth-apps/#available-scopes)) to fetch the list of users you're associated with. (Today, Teletype  uses a scopeless token and has no access to your private information.)
+    - Some users belong to organizations with thousands of members. Just because you're a member of the same organization as someone else doesn't mean that you're comfortable seeing portal invitations from them.
```
dmleong: 🙏

doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md
dmleong: I am concerned about showing this to all windows, as a security precaution. If I'm signed in to multiple computers, I would only want to receive the invitation on the active window

simurai:
@dmleong and @benbalter thanks for all the feedback. ❤️🙇 Instead of spreading out to the many comments, here a proposal based on that feedback. There are a few no's to this RFC, but it also makes the scope smaller, which might not be a bad thing.

- **No presence**. https://github.com/atom/teletype/pull/344#discussion_r175910561
  - We can consider adding it at a later point and offer the option to "appear offline".
- **No option to automatically join**. https://github.com/atom/teletype/pull/344/files#r175908499 Guests will always have to confirm by clicking "Join portal" from the notification. 
    - That option might not even be _that_ practical. Like if I'm using Atom and then all of a sudden a new file/buffer opens without warning, that would feel quite disruptive. I also have no choice of choosing which Atom window I want to use to collaborate or first open a new one.
- **No rejected status**. https://github.com/atom/teletype/pull/344#discussion_r175911322
    - If a guest isn't joining, the host can ask over chat etc. for the specific reason.
- **Option to revoke an invitation** https://github.com/atom/teletype/pull/344/files#r175816061
  - The host can right-click on an invitee and choose "Cancel invitation". Handy when inviting someone by accident.
  image
- **Option to "unfriend" from an invitation** https://github.com/atom/teletype/pull/344#discussion_r175820990
  - You can always remove a collaborator by right-clicking on the past collaborators list. But offer the option at the time when it's probably needed most, might be a good idea. Bellow a possible mockup. The copy could be a bit friendlier and less scary sounding.
  ii notification

Some more thoughts:

- **Link to GitHub profile**
  - In the notification, the host's username is a link to their GitHub profile. So you can first see who it is, if you don't recognize the username.
  - Same for the list, usernames are links.
- Option to **remove all collaborators**
    - This would be handy in case your portal URL gets shared in public and you end up having tons of new collaborators that just clutter up your list and it would be a pain to remove them one by one. Of course, you then are forced to invite by URL again.
  image
- **Collaborators expire**. I think I mentioned this already, but the idea is that if you don't collaborate regularly, that person would be removed from your list automatically. Maybe after 3 months or so. That should keep the list always up to date with only the people you keep collaborating.
- Option to **snooze an invitation**
  - An invitee might be currently busy, but wouldn't mind joining a portal a bit later. So instead of having to keep the notification open, they could snooze it and it would re-appear after 10 min again.
  - A possible workaround could be: Open a new (empty) window where the notification would appear too. Then the notification where the invitee is currently working can be closed.

Still unsure:

- **Be able to block someone**
  - If I remove a collaborator, the only way that collaborator can invite me again is by sending me a URL. And since that has to happen through chat, email, twitter etc., I could and should block that person there instead.
- **Collaborator requests**
  - When a guest joins with a URL, we could **not** automatically add that guest to the past collaborators list. Instead, before the portal is closed, either, host or guest could send a "friend request". Then once accepted, they would be added to the collaborators list. The benefit would be that guests can join portals without being educated or worried about getting invited in the future. So the list of collaborators is a bit more curated and based on mutual consent and not on "one time" collaborations.

dmleong:
@simurai thank you so much for your thoughtful replies! It warms my cold, deadened heart to see such care and thought into making sure Atom is a safe experience for all ❤️ 

- No presence. #344 (comment)
- No option to automatically join. 
- No rejected status. #344 (comment)
- Link to GitHub profile
- Collaborators expire
- Option to snooze an invitation

Love these ☝️  thank you for summing it up so nicely

> Option to "unfriend" from an invitation

Might I also suggest also allowing the ability to block from the invitation? We do this with repository invitations 

repo invites

> Option to remove all collaborators

I appreciate the sentiment behind protecting a user if a portal link is made public ❤️ but  I wonder if this is necessary if we are being more explicit about who you are able to add as a collaborator? If we are to implement the **Collaborator requests** feature you suggested, I don't know if this feature is strictly necessary for anything other than quick cleanup.

> Should Teletype keep its own list of blocked users, or should it be tied to a host's GitHub account? 

If we are using GitHub accounts, it makes sense to me to make sure the two blocked lists are synced. One huge issue our team deals with is data inconsistency with blocking behavior, so I would love it if the two were connected 🙏 

> After the host clicks on the "Share" toggle

A slight visual nitpick (please feel free to ignore this) the toggle feels strange. It feels more in line with sharing active state rather than sharing a link. Perhaps a button or "Open sharing portal"? 

> Ignore this invitation by clicking the X

Blocking info and links to docs about what blocking does would be :100: here!

> Once sharing is over, the guest shows up in the past collaborator list.

I would like the option to add the person to a "permanent" collaborator list rather than them showing up automatically 

> Should the guest also disappear from the host's list? Or can the host still click the "Invite" button, but the notification just doesn't show up for the guest anymore?

If a guest blocks a host, we should 1) remove the guest from the host list 2) not tell the host why. Out of sight, out of mind!

Thanks again for putting such thought and care into this feature!

simurai:
> @dmleong I would like the option to add the person to a "permanent" collaborator list rather than them showing up automatically

Yeah, initially it seemed like a simple and elegant solution to just combine "inviting by URL" and populating the past collaborator list. But separating the two seems the safer route. 👍  And there is also something nice about inviting or getting invited with a URL and know there are no strings attached.

Ok, so here two variations on how populating the collaborators list could work.

### 1. While collaborating

It's pretty similar to the [walkthrough](https://github.com/atom/teletype/pull/344#issuecomment-374869948) above, but now adding a collaborator to the list needs to be done manually. In the mockup they are called "trusted collaborator", just to give it a name.

1. Click on a collaborator's avatar, then click on the "Send Request" button.
2. A dialog appears where the request can get accepted (or not).

image

Hmm.. not sure, I can see some flaws. The biggest is that a request can only be sent while the collaboration is still ongoing. So people would have to keep remembering to send a request before the portal gets closed.

### 2. Any time

In this version, no portal needs to be open and it's not related to "inviting by URL" in any way. It's close to "Friend requests" on Facebook.

- Above the collaborators list is an input that filters as you type a name.
- But if you paste a GitHub username and press enter, it looks up if that username exists and shows it underneath with an avatar. There should probably be a hint about it when filtering returns no matches. Or could also be part of the input placeholder.
- Now you can click on "Send request".

image

- Next time the other person opens Teletype's popover, there is a counter next to "Requests", hinting that new requests are waiting.
- After switching to the "Requests" tab, new requests will be shown. Ready to be accepted or ignored.

image

I really like the non-aggressiveness of this approach. There are no dialogs popping up, no invite emails sent, just the blue counter. You can easily ignore it and get back to it whenever you feel like. If it's urgent and you can't wait, just ask them to accept your request over chat or so.

dmleong:
😍 Any time requests 😍 

One question: if a user doesn't have a GitHub account (strange, I know) what is the fallback @simurai?

simurai:
> if a user doesn't have a GitHub account (strange, I know) what is the fallback

No fallback! :grimacing: After installing the Teletype package, you get asked to authenticate and in case you're not signed into GitHub, you'll get this login form where you'd have to "Create an account" first.

image

Requiring a GitHub account is be a bit of a barrier, but at the same time collaborators can be sure about people's identities.. well, minus the impersonators.

dmleong:
>  you'd have to "Create an account" first.

Awesome! Glad that loophole is pretty cleanly closed. What's the flow if an invitee has blocked the inviter?

jasonrudolph:
@damieng @benbalter: Thank you for the thoughtful feedback. :bow:💟

@simurai: Thank you thinking through our options for incorporating this feedback and for iterating on the UX. :bow::sparkles:

I'm hoping to circle back to this discussion :soon: to incorporate these updates into the RFC and then open it up for another round of review.

simurai:
> What's the flow if an invitee has blocked the inviter?

AFAIK, Teletype doesn't check against GitHub's blocked user list. It only uses GitHub to get a name and avatar. So blocked users can join if they know the Portal URL.

We could show a "You can't join because you got blocked by the host", but that needs some careful wording to not make people angry. Another question would be: What if another guest blocked me but not the host, can I still join? It's probably strange, seeing some sort of 👻 ghost type without a name and avatar. Or it could just warn me: "Hey, watch out. A user that you blocked is in here". And then I can decide if I want to join or not.

Should we open another issue? How blocked users should be handled probably needs its own RFC.

benbalter:
> Thank you for the thoughtful feedback

Thank YOU for reaching out. That's what we're here for. Hope the feedback was helpful. Sounds like things landed in a good place, but we're here to help any time.

dmleong:
> AFAIK, Teletype doesn't check against GitHub's blocked user list. It only uses GitHub to get a name and avatar. So blocked users can join if they know the Portal URL.

If Teletype is using the GitHub API, it shouldn't be a huge lift to [check if you are blocking someone](https://developer.github.com/v3/users/blocking/#check-whether-youve-blocked-a-user)

> We could show a "You can't join because you got blocked by the host",

We tend to default to "You can't do this at this time" to make it as innocent as possible.

>  Or it could just warn me: "Hey, watch out. A user that you blocked is in here". And then I can decide if I want to join or not.

This seems like the best ☝️ 

Thanks again for reaching out to us! Let us know if you have any other questions and I'm really happy to see this working out well!

jasonrudolph:
@dmleong @benbalter: :wave: Hi again! I've updated the proposal in a way that I think will offer _most_ of the product benefits we're seeking (i.e., easier ability to quickly collaborate with coworkers and friends) while also incorporating all our your thoughtful feedback. When you have some time, can you please review the updated proposal and share any feedback you might have from a community and safety perspective?

- [Diff](https://github.com/atom/teletype/pull/344/commits/24a8cd1)
- [Rendered RFC-004](https://github.com/atom/teletype/blob/24a8cd1/doc/rfcs/004-quickly-collaborate-with-coworkers-and-friends.md)

Thanks for all your help with this. 💟:bow:

---

Note: 8863d0bdb13534ad50770df0ca5253f41903b9a2 split the original RFC into two separate RFCs: RFC-003 and RFC-004:
- Since all of the review discussion above focused on the behavior of the "buddy list", that topic now lives in RFC-004 (linked above), and we'll iterate on that RFC based on your feedback.
- RFC-003 simply covers replacing the previous use of portal IDs with more user-friendly portal URLs. That noncontroversial change :innocent: shipped yesterday in [Teletype 0.11](https://blog.atom.io/2018/03/29/new-in-teletype-faster-setup-and-improved-multi-file-support.html). Feel free to ignore RFC-003.

dmleong:
@jasonrudolph as per our 📹 call earlier this week, I'm happy with how this landed! Thanks so much for all your work on this!

simurai:
Here a mockup with a "Trusted" and "Recent" list:

screen shot 2018-04-06 at 11 51 20 am

After clicking on "Trust", they would appear in the "Trusted" list. From there they can be invited directly into your portal (if they trust you too).

Hmm.. not sure. Somehow I can't help thinking it's only the "best of the bad options". :laughing: Still better than nothing, but no matter how well we explain it, I think it will still cause confusion.

simurai:
More ideas that try to avoid "sending invites that may never show up":

### 1. Change permissions

3 out of the 5 [alternative approaches](https://github.com/atom/teletype/blob/24a8cd1/doc/rfcs/004-quickly-collaborate-with-coworkers-and-friends.md#rationale-and-alternatives) mention "Users would need to grant Teletype substantially greater permissions" as the show stopper. In our last design review, @broccolini made this suggestion:

1. Initially ask for the [Public data only](https://user-images.githubusercontent.com/2988/38115271-65e65928-3379-11e8-9945-4a15ceb857fe.png) permission. 
2. But when you want to add someone to your collaborator list, ask for the [Personal user data](https://user-images.githubusercontent.com/2988/38115272-65f69784-3379-11e8-8753-4da6547c7edb.png) permission.

Teletype would only ask for the greater permissions when really needed. So that people that get turned off by the greater permissions, could still keep using Teletype with URLs.

**Concerns**: It might not be easy to "switch/upgrade" permissions and a user would have to manually revoke access first? That would be kinda a cumbersome process to explain.


### 2. Regular collaborators

- It's pretty much the same as with the very first "past collaborators" approach, except that you only end up on each other's regular collaborator list after collaborating least **3 times** within 3 months (numbers may be adjusted).
- Also this only happens between hosts and guests, but not between two guests that were in a portal together.
- So if you share your portal URL in a place where someone can join 3 times in a short period, we can be pretty certain that they wanna keep collaborating in the future.

**Concerns**: Even if it's much harder for a bad actor to end up on a collaborators list, it's not impossible.

### 3. Authentication codes

- This is similar to the previous "invite with portal ID", but used as authentication for adding someone to the collaborators list.
- There is a "add collaborator" button, when clicked it generates an ID that can be copied and pasted into a chat like Slack.
- People can then copy the ID, also click on "add collaborator" and paste the ID into the text input. After they appear on the collaborators list and can be invited directly.
- IDs expire after a certain time (10-15min). It's more like these 2FA codes shown in apps like Google Authenticator. So no worries that someone will find an old ID somewhere that is still valid.
- Another nice thing is that you can share an ID with multiple people (in the same chat room), for example after joining a new team.
- Why not use URLs too? To not be mistaken with "invite to a portal with a URL". Also by having to paste an ID in the right place, we make sure people realize what they are doing.

**Concerns**: None! :raised_hands: Well, except that it takes some extra effort to have less effort later.

---

Do any of ☝️ these sound better?

dmleong:
@simurai I'm a little confused about some of the notes you've left here, but here's my thoughts. 

> Even if it's much harder for a bad actor to end up on a collaborators list, it's not impossible

A harasser, more often than not, is someone who is _familiar to the target_. Think an ex of yours who goes through and likes all your instagram posts for the past year just to let you know that you can't get rid of them. What we're trying to optimize for is *revoking consent at any time*. 

I think the "losing a pending invitation" issue is less of a concern than the ability to clearly define who is a trusted person and being able to revoke that privilege. Tagging @katmeister for some more nuanced design perspectives here

jasonrudolph:
> Somehow I can't help thinking it's only the "best of the bad options". 😆 Still better than nothing, but no matter how well we explain it, I think it will still cause confusion.

@simurai: I found myself feeling that way as well. After expressing that concern to @dmleong, we had a call last week, and she helped me identify some ways that we can improve this proposal to provide better UX for trusted collaborators. I intend to update the proposal to incorporate those improvements, but it might be next week before I can get to those updates in place. Stay tuned. 🙂

jasonrudolph:
> After expressing that concern to @dmleong, we had a call last week, and she helped me identify some ways that we can improve this proposal to provide better UX for trusted collaborators. I intend to update the proposal to incorporate those improvements, but it might be next week before I can get to those updates in place.

98cd3d10e1992266525c9a7e603f6094b134b3b2 updates the RFC to incorporate the improvements that came out of my discussion with @dmleong. 

👉👉👉 [Rendered RFC](https://github.com/atom/teletype/blob/9981c551565cfdf1fb6709cabd95978c90822d36/doc/rfcs/004-quickly-collaborate-with-coworkers-and-friends.md#quickly-collaborate-with-coworkers-and-friends) 👈👈👈

@simurai: Would you be willing to look over the updated RFC and offer some mockups? If you search for "TODO" in the RFC, you'll see 4 areas where I think we'd benefit from a new or updated mockup. And of course, if you have any feedback on the RFC, I certainly welcome it. 😃

simurai:
> You can select a trusted collaborator that is currently online

Great idea to bring the "online status" back into the mix. I guess a downside it that you have to wait until someone is online before you can invite them, but with all the considerations we went through, I think an ok compromise. 👍

I added a few mockups. We might still wanna change a few things, but it shouldn't affect how things work.

doc/rfcs/004-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,168 @@
...
+
+People will collaborate less often. Given the additional steps needed to start collaborating, there will be more instances where people decide that it's not worth the effort (i.e., people will enjoy rich collaboration less often).
+
+## Unresolved questions
+
+##### What unresolved questions do you expect to resolve through the RFC process before this gets merged?
+
+- If I invite a trusted collaborator to join my portal, and that person has multiple Atom windows open, can we show the invitation only in the frontmost window?
+- If I invite a trusted collaborator to join my portal, and that person has Atom windows open on multiple computers, are we OK with showing the invitation on each computer?
+- Where will the lists of recent and trusted collaborators appear in the UI? How do I show/hide these lists?
```
simurai: The lists appear in the "popover" when clicking the Teletype icon in the status bar.

I was debating if they should be two separate lists that have tabs, but that will probably have the effect that you add less people as trusted collaborators because you have to switch between them. With both lists on top of each other and scrollable, you more likely discover newly added people in the recent collaborator list.

doc/rfcs/004-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,168 @@
...
+- In order to provide the functionality described above, does teletype-server need to persist your list of trusted collaborators, or can we meet these needs while only storing this data locally?
+- If the host invites a trusted collaborator to their portal and then closes the portal before the invitee joins the portal, what should an invitee see/experience to inform them that the portal no longer exists?
+
+##### What related issues do you consider out of scope for this RFC that could be addressed in the future independently of the solution that comes out of this RFC?
+
+See [out of scope](#out-of-scope) section above.
+
+---
+
+[A] A user is **online** if they have Atom open, and Teletype is installed, and they are signed into Teletype, and their network allows them to successfully communicate with `api.teletype.atom.io`. In all other situations, the user is considered to be **offline**.
```
simurai: > their network allows them to successfully communicate with `api.teletype.atom.io`.

What about if the computer goes to sleep, will a user be shown as offline?

jasonrudolph: Yes. If your computer goes to sleep, we'll want Teletype to show you as offline.

jasonrudolph:
> I added a few mockups. We might still wanna change a few things, but it shouldn't affect how things work.

Thanks, @simurai. :bow:

I agree that we'll probably want to iterate on the design and the copy as we build this out, but I think these mockups will meet our needs for the RFC. Thank you! :zap:

doc/rfcs/004-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,166 @@
...
+
+1. Host selects a person in their list of trusted collaborators
+2. Host invites the selected person to a portal, and Atom automatically shares a portal for the host and invites the selected person
+3. Guest accepts invitation to join portal, and Atom joins the portal
+
+## Explanation
+
+### Accept portal invitations from a trusted collaborator
+
+Teletype provides a list of your recent collaborators. Each time a guest joins your portal, Teletype adds them to your list of recent collaborators. Each time you join a portal, Teletype adds the host to your list of recent collaborators.
```
dmleong: Can we make sure that we do not lose the ability to remove recent collaborators? Removing trusted collaborators is noted before, but it would be great to also note how to remove recent collaborators!

doc/rfcs/004-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,166 @@
...
+
+If you have designated a person as a trusted collaborator, and they have also designated you as a trusted collaborator, then you can see whether they're online [A], and they can see whether you're online.
+
+You can select a trusted collaborator that is currently online (an "invitee" in this context) and invite them to join your portal. If you're not already hosting a portal, Teletype automatically creates a portal for your workspace.
+
+![](https://user-images.githubusercontent.com/378023/38798163-c1dc4226-419b-11e8-9b96-5db9a9c90221.png)
+
+If the invitee chooses to join the portal, the portal opens in the invitee's workspace (just as it does when joining a portal via a URL).
+
+If the invitee declines to join the portal, Teletype notifies you that the guest declined your invitation. You cannot re-invite that invitee to the same portal. (There's probably a reason why they didn't join, so there's no need to bug them even more. If you think they declined by accident, reach out to them via a third-party chat service and send them the URL to join your portal.)
```
dmleong: ~~I would prefer that Teletype does not inform you that a guest has declined your invitation, as part of the "negative feedback loop" I mentioned in 📹 . Harassers often view these "negative" events as a badge of honor (i.e. some people celebrate being blocked on Twitter as a sign they got under their target's skin).~~

~~If you are already relying on a third party chat application, I don't think it's necessary to send a "decline" back to the host.~~

EDIT: I didn't notice that this is in the "trusted collaborator" section. I think it would be important to distinguish that we ONLY do this for trusted collaborators, or have the ability to have pre-filled reasons to decline an invitation (i.e. "Sorry, I'm busy right now") or an away message (which is outlined as out of scope below). I think it would be important to clarify we do not send back a decline message if the two users are not mutually trusted collaborators.

doc/rfcs/004-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,166 @@
...
+
+    ![](https://user-images.githubusercontent.com/378023/38803217-79518b02-41a9-11e8-9f07-0f0085fc994f.png)
+
+- At any time, you can right-click a person in your list of trusted collaborators and remove them:
+
+    ![](https://user-images.githubusercontent.com/378023/38803345-e0087c3e-41a9-11e8-8c59-23c5a9a38bb9.png)
+
+If you decide that you want to collaborate with them again, you can send them a URL (via Slack, IRC, etc.) to join your portal, or you can ask them to send you a URL (via Slack, IRC, etc.) so that you can join their portal, or you can add them back to your list of trusted collaborators.
+
+## Additional privacy and safety considerations
```
dmleong: I love this whole section ❤️ 😭

doc/rfcs/004-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -0,0 +1,166 @@
...
+2. **Invite anyone to collaborate by username, and ignore invitations from blocked users**: To reduce the harassment concerns described in the previous bullet, Teletype could require that you grant it additional permissions so that it can read your [blocked user settings](https://developer.github.com/v3/users/blocking/) to determine whether to show you an invitation from a particular user. With this approach, if a user blocks a person on github.com, Teletype could automatically prevent invitations from the blocked person. However, this approach introduces tradeoffs that we'd prefer to avoid:
+    - Users would need to grant Teletype substantially greater permissions in order for Teletype to check users' block lists. Due to coarse-grained OAuth scopes, users would need to [grant Teletype permission to read/write their private email addresses, update their github.com profile, etc.][user-scope-request] Teletype would no longer be able to simply require [read-only permission to each user's public info][scopeless-request].
```
dmleong: I really appreciate this call out and concern for permissions, thank you 🙇‍♀️

doc/rfcs/004-quickly-collaborate-with-coworkers-and-friends.md
```diff
@@ -145,8 +145,7 @@ People will collaborate less often. Given the additional steps needed to start c
 
 ##### What unresolved questions do you expect to resolve through the RFC process before this gets merged?
 
-- If I invite a trusted collaborator to join my portal, and that person has multiple Atom windows open, can we show the invitation only in the frontmost window?
-- If I invite a trusted collaborator to join my portal, and that person has Atom windows open on multiple computers, are we OK with showing the invitation on each computer?
+- If I invite a trusted collaborator to join my portal, and that person has multiple Atom windows open, can we show the invitation only in the frontmost window? Similarly, if that person has Atom windows open on multiple computers, are we OK with showing the invitation on each computer? _Answer_: The invitation will appear in all open Atom windows where the user is signed into Teletype. Once the user accepts or declines the invitation on in one Atom window, Teletype will dismiss the invitation in the other Atom windows. [[discussion](https://github.com/atom/teletype/pull/344#pullrequestreview-105123633)]
```
simurai: I think that's a good solution. It also helps in another scenario: When you wanna "snooze" an invitation. If you get invited and want to join, but still need to spend a few minutes finishing up something, then the invitation dialog blocks your editor. So in this case, you could just open a new window (where Teletype shows the invitation too), then click on the X of the current window. This will only close the dialog, but keep the invitation active (in the other window).

doc/rfcs/004-quickly-collaborate-with-coworkers-and-friends.md
jasonrudolph: > I think it would be important to clarify we do not send back a decline message if the two users are not mutually trusted collaborators.

If the two users are not mutually-trusted collaborators, they can't send invitations to each other at all, which means that there's no invitation to decline. 😇

doc/rfcs/004-quickly-collaborate-with-coworkers-and-friends.md
dmleong: @jasonrudolph  Sure! Let's say I accidentally click on a link in Slack and realize it's someone I don't actually like that's hosting the Teletype portal. That person then shows up in my recent collaborators list. It's not explicitly clear to me if/when their username will disappear from my recent collaborators list, so I would like the ability to click an (x) to remove them from the list. Does that make sense?

doc/rfcs/004-quickly-collaborate-with-coworkers-and-friends.md
dmleong: Gotcha! Thanks for the clarification

doc/rfcs/004-quickly-collaborate-with-coworkers-and-friends.md
jasonrudolph: @dmleong: Thanks for that context. :zap::bow: 

46c56136ab02dd70f707726896000e0ba3928c82 notes that Teletype should provide the ability to remove anyone from your list of past collaborators.

markaduol: If the user opens up new Atom window without any recent (or "favorite") projects loaded, a possible solution may be to have a "Recent Projects" tab holds a menu of the user's recent projects. After clicking on the menu item for "Project X" (for instance) the view will then transition from a menu of projects, to a list of recent collaborators on Project X.

PR #346, opened 2018-03-20
Join portal via URL (Part 1)

jasonrudolph:
### Description

This pull request implements the following item from the task list in https://github.com/atom/teletype/issues/109:

> - [ ] Teach Teletype to join portal via URL
>     - Add URI handler
>     - Add command for getting the URL
>     - Automatically join the portal when URL is followed (for now)
>     - Don't provide any UI for getting the URL (for now)

This is the first in a series of pull requests that will implement the various aspects described in https://github.com/atom/teletype/issues/109. Once all items in that task list have been addressed, we'll publish a new Teletype release with support for joining a portal via a URL.

### Demo

demo

### TODO

- [x] Add basic support for joining portal via URL
- [x] Resolve all TODOs present in the diff

### Verification process

- [ ] Can use `Teletype: Copy Portal Url` command to copy URL when hosting a portal
- [ ] `Teletype: Copy Portal Url` command is not available when not hosting a portal
- [ ] Following valid portal URL launches Atom and joins portal
- [ ] Following the same URL multiple times has no effect
- [ ] Following a malformed URL results in an error notification
    - `atom://teletype/portal/42`
    - `atom://teletype/whoops`
    - `atom://teletype/`
    - `atom://teletype`
- [ ] Following a URL to join your own portal results in an error notification
- [ ] Following a URL to join a closed portal results in an error notification
- [ ] Following a URL to join a portal when signed out of Teletype displays the "sign in" popup

### Applicable Issues

#109 
#344

PR #350, opened 2018-03-21
Join portal via URL (Part 3 of 3)

### LINKED ISSUE #109: Make the portal ID a link that opens in Atom

jasonrudolph:
Closes #109.

### Description of the Change

Building on the work in #346 and #348, this pull request implements the final item from the task list in https://github.com/atom/teletype/issues/109:

> - [ ] Ask the user to confirm before joining the portal [[motivation](https://github.com/atom/teletype/blob/e35483b1fa69c3ef1ef56f0c437000d4dcda934b/doc/rfcs/003-quickly-collaborate-with-coworkers-and-friends.md#join-a-portal-via-url)]

### TODO

- [x] Show modal dialog when user follows portal URL
- [x] Ask @simurai for help styling the dialog
- [x] Clicking outside of dialog should dismiss dialog without joining portal
- [x] `core:cancel` should dismiss dialog without joining portal
- [x] `core:confirm` should join the portal
- [x] In the settings UI, consider grouping the dev-specific options together (similar to the groupings used by the atom-ide package)
- [x] Wire up config option 
- [x] Consider alternative names and descriptions for the config option
- [x] Define and execute a verification process

### Verification Process

- [x] By default, following a URL asks before joining
- [x] Clicking outside of dialog dismisses dialog without joining portal
- [x] Clicking the 'x' dismisses dialog without joining portal
- [x] Hitting <kbd>esc</kbd> dismisses dialog without joining portal
- [x] Clicking "Join portal" joins portal
- [x] Hitting <kbd>return</kbd> joins portal
- [x] After enabling "join without asking" checkbox, following a URL automatically joins the portal
- [x] After re-enabling the "always ask" option in Teletype settings, following a URL asks before joining
- [x] After disabling the "always ask" option in Teletype settings, following a URL automatically joins the portal
- [x] Verify dialog renders properly in the most popular UI themes

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,11 +7,11 @@ module.exports = new TeletypePackage({
   commandRegistry: atom.commands,
   tooltipManager: atom.tooltips,
   clipboard: atom.clipboard,
-  pusherKey: atom.config.get('teletype.pusherKey'),
+  pusherKey: atom.config.get('teletype.dev.pusherKey'),
   pusherOptions: {
-    cluster: atom.config.get('teletype.pusherCluster'),
+    cluster: atom.config.get('teletype.dev.pusherCluster'),
     disableStats: true
   },
-  baseURL: atom.config.get('teletype.baseURL'),
+  baseURL: atom.config.get('teletype.dev.baseURL'),
   getAtomVersion: atom.getVersion.bind(atom)
 })
--- a/index.js
+++ b/index.js
@@ -1,5 +1,6 @@
 const TeletypePackage = require('./lib/teletype-package')
 module.exports = new TeletypePackage({
+  config: atom.config,
   workspace: atom.workspace,
   notificationManager: atom.notifications,
   packageManager: atom.packages,
--- a/lib/host-portal-binding-component.js
+++ b/lib/host-portal-binding-component.js
@@ -69,7 +69,7 @@ class HostPortalBindingComponent {
         ),
         $.div({className: 'HostPortalComponent-connection-info-portal-url ' + statusClassName},
           $.input({className: 'input-text host-id-input', type: 'text', disabled: true, value: this.getPortalURI()}),
-          $.button({className: 'btn btn-xs', onClick: this.copyPortalIdToClipboard}, copyButtonText)
+          $.button({className: 'btn btn-xs', onClick: this.copyPortalURLToClipboard}, copyButtonText)
         )
       )
     } else {
@@ -91,7 +91,7 @@ class HostPortalBindingComponent {
     }
   }
 
-  copyPortalIdToClipboard () {
+  copyPortalURLToClipboard () {
     const {clipboard} = this.props
     clipboard.write(this.getPortalURI())
 
--- a/lib/join-via-external-app-dialog.js
+++ b/lib/join-via-external-app-dialog.js
@@ -1,0 +1,68 @@
+const etch = require('etch')
+const $ = etch.dom
+const {TextEditor} = require('atom')
+
+module.exports =
+class JoinViaExternalAppDialog {
+  constructor ({config, commandRegistry, workspace}) {
+    this.commandRegistry = commandRegistry
+    this.workspace = workspace
+    this.confirmOnce = this.confirmOnce.bind(this)
+    this.confirmAlways = this.confirmAlways.bind(this)
+    this.cancel = this.cancel.bind(this)
+    etch.initialize(this)
+    this.disposables = this.commandRegistry.add(this.element, 'core:cancel', this.cancel)
+  }
+
+  destroy() {
+    if (this.panel) this.panel.destroy()
+
+    this.disposables.dispose()
+    etch.destroy(this)
+  }
+
+  async show (uri) {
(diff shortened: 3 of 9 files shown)


jasonrudolph:
> - [ ] Ask @simurai for help styling the dialog

@simurai: Would you have time today to look into styling the modal dialog introduced in this PR?

To trigger the dialog locally, you can check out this branch, link the package, and follow a portal URL from another app. For example, inside the terminal, you can run:

    open atom://teletype/portal/00000000-0000-0000-0000-000000000000

The relevant code for the dialog is here: https://github.com/atom/teletype/blob/3d528c2fa6fcbd3dccd5775cea820eade6196b2c/lib/join-via-external-app-dialog.js#L61-L75

I think it makes sense to style it similar to your mockup from https://github.com/atom/teletype/pull/344#issuecomment-374869948:



For the scope of _this_ PR, we won't yet be showing the host's username, but we can add that in a follow-up PR.

If you do have time to dig into this, please feel free to alter the structure of the HTML and add CSS classes as you see fit. :bow:

simurai:
Added a "Ask before joining a portal" checkbox: 

dialog

An alternative would be to disable it and change the wording to something like "Join portals without asking":

dialog 2

I think that would better and only have one blue "thing", but using the negative (without) wording isn't as clear. Maybe something else will come to mind.

jasonrudolph:
> An alternative would be to disable it and change the wording to something like "Join portals without asking":
>
> dialog 2
> 
> I think that would better and only have one blue "thing", but using the negative (without) wording isn't as clear. Maybe something else will come to mind.

@simurai: Thanks for sharing this idea. Building on that, I've updated the dialog to disable the checkbox by default, and I've updated the copy to hopefully provide more clarity:

jasonrudolph:
@simurai: I tested the dialog in a bunch of popular UI themes, and it looks good in most of them. 😅  There are just a couple themes that had issues. Can you take a look at these and let me know how you think we should proceed?

### [nucleus-dark-ui](https://atom.io/themes/nucleus-dark-ui)

For some reason, the text doesn't show up in this theme:

screen shot 2018-03-22 at 16 59 37 pm

Other modals like the fuzzy finder work fine in the nucleus-dark-ui theme, so I'm not sure why this new Teletype dialog doesn't render correctly. Do you think this is a bug in the nucleus-dark-ui theme, or are there changes we should make to Teletype?

### [city-lights-ui](https://atom.io/themes/city-lights-ui)

The portal URL wraps in this theme:

screen shot 2018-03-22 at 17 09 02 pm

Should we adjust the dialog to prevent the portal URL from wrapping?

simurai:
I like the new copy and made some changes since it's longer.

- [x] Bigger join button
- [x] The checkbox label wraps nicer -> lines are aligned.

portal

> Do you think this is a bug in the nucleus-dark-ui theme

Yeah, I think it's something nucleus-dark-ui should fix. We _could_ fix it in the Teletype package, but that doesn't fix the root cause and other modals might be broken too. I'll make a PR there.


> Should we adjust the dialog to prevent the portal URL from wrapping?

I think it's ok that the URL wraps. We could just cut it off, but wrapping makes it easier to copy&paste in case you decide to use a different Atom window.

copy-pasta

jasonrudolph:
Thanks for the styling updates, @simurai. :bow::zap:

> Yeah, I think it's something nucleus-dark-ui should fix. ... I'll make a PR there.

You're a superstar! 😻

PR #351, opened 2018-03-22
Fix stacking of link icon

### LINKED ISSUE #345: Icon not displayed correctly when following another user
After the improvements introduced in #332, it seems like the little "link icon" displayed next to the user I am following isn't being rendered correctly anymore:

screen shot 2018-03-20 at 09 07 33

I assume the icon should go in front of the blue background, but for some reason I think `z-index` isn't working correctly. I have noticed we're using pseudo elements to represent the background and the icon link:

screen shot 2018-03-20 at 09 08 06

Maybe there's a way of doing everything in the same element or swap the rendering order somehow without using `z-index` (or maybe by creating a new [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context))?.

This is the configuration I am using:

* UI theme: One Dark UI
* Syntax theme: One Dark Syntax
* Atom version: 1.27.0-dev-079a43f7b (but I can reproduce on 1.25 stable too)
* Teletype version: 0.10.0
* Relevant personal style sheet configuration: none
* Relevant configuration settings: none
* Can reproduce in safe mode: yes

/cc: @simurai @jasonrudolph

simurai:
### Description of the Change

This PR tries to fix #345 by adding more `z-index`es.

### Alternate Designs

Use a separate element for the background.

### Benefits

Icon is not hidden

### Possible Drawbacks
None

### Verification Process

- [x] Join a portal.
- [x] Click on a user's avatar.
- [x] Link icon should be visible.

### Applicable Issues

Closes #345

### DIFF
--- a/styles/teletype.less
+++ b/styles/teletype.less
@@ -384,13 +384,15 @@
       pointer-events: none;
 
       &::before {
-        z-index: 1;
+        position: relative;
+        z-index: 2;
       }
 
       // background cover
       &::after {
         content: "";
         position: absolute;
+        z-index: 1;
         left: -2px;
         top: 4px;
         width: 15px;


simurai:
@as-cii I still can't reproduce it, but I added another `z-index` to the 2nd pseudo element. Let me know if that helped.

as-cii:
Yup, this solves it for me! Thanks @simurai!

PR #353, opened 2018-03-22
[Code Submission] RFC-002: Sync Buffer Path Changes from Host To Guest

CaptainJohnyAppleSeed:
### Description of the Change

This is the changes to the **Teletype** package mentioned in [RFC-002 Tracking](atom/teletype/352).

Please see [RFC-002](https://github.com/atom/teletype/issues/352) for details about the change.

### Verification Process
We have made 2 tests in Buffer-Binding, we are open to making more tests if needed.

### Applicable Issues
Original Issue: [#147](atom/teletype/147)

Issue #352, opened 2018-03-22
[Tracking] RFC-002: Sync Buffer Path Changes from Host To Guest


CaptainJohnyAppleSeed:
We intend to implement [RFC-002](https://github.com/atom/teletype/blob/master/doc/rfcs/002-sync-buffer-path-changes-from-host-to-guest.md) through 2 pull requests which need to be implemented simultaneously in **teletype** and **teletype-client**

Refs: [Original Issue #147](atom/teletype/147)

CaptainJohnyAppleSeed:
Closing atom/teletype-client#55 in favor of the implementation in atom/teletype-client#57 
Refactored #353 to reflect this change.

CaptainJohnyAppleSeed:
Next step is to update real quick RFC-002 with the updated workflow

CaptainJohnyAppleSeed:
@jasonrudolph Thanks so much for working with us for this RFC! I believe that we are now all set, and this tracking issue can be closed!

PR #356, opened 2018-04-03
Update Workflow

CaptainJohnyAppleSeed:
Reflect updated workflow in atom/teletype-client#57 and atom/teletype#353
Refs: #352


CaptainJohnyAppleSeed:
@jasonrudolph we updated the RFC-002 workflow.

PR #399, opened 2018-07-31
[do not merge] Test CI against atom/ci changes

smashwilson:
Ensure that the changes in atom/ci#86:

* Do not break the stable and beta channel builds in Travis (which are using the older npm version)
* Fix the dev channel build in Travis (which is using the newer npm version)


smashwilson:
👌

PR #414, opened 2018-09-21
Toggle collaborators display

AdrienHalletUCL:
### Description of the Change

The package's settings page now has a new boolean allowing to display or hide the collaborators avatars in the text editor.

### Alternate Designs

This change is the most simple one. Putting this option as a setting does not overload the user interface. I considered adding it as a menu item but I don't think it is an option used enough to be displayed as such.

### Benefits

As seen in the poll, most people use teletype with only few participants. My collaborators and I mostly work in group of two or three and in the same workspace. Our utilisation of the package does not require any knowledge of each user's position. We often use the plugin during meeting to take collaborative notes on half a laptop's screen. The reduced window size puts the avatars over the code, and the blinking colors around the avatars is distracting. Allowing it to simply hide those avatar solves all these problems.

### Possible Drawbacks

I don't see any.

### Verification Process

I followed the official contribution guide. After implementing this feature I ran the tests (but did not add any) wich still passed. I ran local manual tests to check the correct behaviour and display of the avatars.

### DIFF
--- a/lib/guest-portal-binding.js
+++ b/lib/guest-portal-binding.js
@@ -243,7 +243,7 @@ class GuestPortalBinding {
   didChangeActivePaneItem (paneItem) {
     const editorProxy = this.editorProxiesByEditor.get(paneItem)
 
-    if (editorProxy) {
+    if (editorProxy && atom.config.get('teletype.displayCollaboratorsOnScreen')) {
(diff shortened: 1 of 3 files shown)


AdrienHalletUCL:
It is a bit weird that the tests worked locally but not here. Any help appreciated.

PR #429, opened 2018-11-27
Include package-lock.json

smashwilson:
As of apm 2.0.0 (included in Atom 1.31.0), apm now generates lockfiles with `apm install` and respects them with `apm ci`, so we should commit these now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
 node_modules
-package-lock.json
 .env
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,4027 @@
+{
+  "name": "teletype",
+  "version": "0.13.3",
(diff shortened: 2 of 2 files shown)


jasonrudolph:
@smashwilson: Will we need to update https://github.com/atom/ci in order to have CI start respecting these lockfiles? I don't _think_ https://github.com/atom/ci currently uses `apm ci`.

smashwilson:
We won't _need_ to, but it'd be nice to. The tricky part is doing this in a way that doesn't break for CI setups that are running against older Atom versions (... assuming there are some?). But I think that might be as easy as doing:

```sh
${APM_SCRIPT_PATH} ci || ${APM_SCRIPT_PATH} install
```

PR #437, opened 2019-02-05
Add libdbus-1-3 dependency for Electron 3 builds of Atom

daviwil:
### Description of the Change

Latest Teletype master builds on Travis CI against Atom Dev are failing with a renderer crash.  One obvious starting point is this error that gets written out in the `dev` build log:

```
/home/travis/atom/usr/share/atom-dev/atom: /lib/x86_64-linux-gnu/libdbus-1.so.3: no version information available (required by /home/travis/atom/usr/share/atom-dev/atom)
```

This change adds `libdbus-1-3` to the `apt-get install` call to ensure that this dependency exists on the CI machine.

### Alternate Designs

None at the moment, let's see if this fixes the issue.

### Benefits

Passing `dev` builds on `master`.

### Possible Drawbacks

None.

### Verification Process

- [x] Verified that no renderer process crash occurs locally on my Linux dev machine
- [ ] Green CI build after adding the new dependency

### Applicable Issues

None.


daviwil:
This error is a red herring and isn't a cause for concern: libdbus-1.so.3: no version information available.

Continuing to investigate...

PR #463, opened 2019-07-01
Fix tests that run against Atom dev channel by upgrading to Ubuntu 16

nathansobo:
Refs https://github.com/atom/ci/issues/93

We recently [switched from Travis to Azure Pipelines for our Atom Linux artifacts](https://github.com/atom/atom/pull/19597), and as an unexpected consequence, package builds that rely on Atom started failing with the following error installing the Atom package:

```
Downloading latest Atom release on the dev channel...
dpkg-deb: error: archive 'atom-amd64.deb' has premature member 'control.tar.xz' before 'control.tar.gz', giving up
```

On Travis, we were building Atom on Ubuntu 14. On Azure Pipelines, we're using Ubuntu 16, and it seems that packages built on Ubuntu 16 trigger this error on Ubuntu 14, which is the version currently used on our package builds as well. Seeing as how Ubuntu 14 has reached "end of standard support", I think it makes sense to upgrade our packages to run on newer versions rather than try to move backwards with our main Atom build.

It seems that Atom 1.38 (which runs on Electron 2.x) needs an additional dependency to be installed which isn't available on the Ubuntu 16 image. Versions of Atom greater than 1.38 seem to build fine without that dependency, so I've added a comment on the dependency so we can remove it in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: node_js
 node_js: '7'
 sudo: false
 os: linux
-dist: trusty
+dist: xenial
 
 env:
   matrix:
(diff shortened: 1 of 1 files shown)


Arcanemagus:
> It seems that Atom 1.38 (which runs on Electron 2.x) needs an additional dependency to be installed which isn't available on the Ubuntu 16 image.

Interesting, I actually added `libgconf-2-4 (>= 3.2.5) | libgconf2-4` as a package dependency to Atom in https://github.com/atom/atom/pull/19358, however shortly after that the Electron 3 PR was merged which means that dependency can be removed again since it is only required for Electron 2.

Arcanemagus:
Ah ha, I just realized you probably meant "isn't available **by default**". That PR is only in v1.39 betas right now.

nathansobo:
Yeah, I think we can remove that package installation once 1.39 goes to stable.

Issue #514, opened 2020-03-19
Atom/teletype electron 5 fix


darangi:
Electron API changed when we upgraded from 4 to 5, it returns a different object structure, while electron 5 is still in nightly, we need to maintain both object structures to pass the tests and we will delete the extra/unnecessary tests once electron 5 reaches stable.
### Fix: https://github.com/atom/teletype/pull/515

darangi:
These are logs from the CI 
```
 1) EditorBinding relays local selections and creates cursor decorations on the local editor based on the remote ones:
      AssertionError [ERR_ASSERTION] [ERR_ASSERTION]: Expected values to be loosely deep-equal:
[
  {
    head: Point {
      column: 4,
      row: 11
    },
    tail: Point {
      column: 0,
      row: 10
    }
  },
  {
    head: Point {
      column: 0,
      row: 20
    },
    tail: Point {
      column: 5,
      row: 20
    }
  },
  {
    head: Point {
      column: 2,
      row: 4
    },
    tail: Point {
      column: 0,
      row: 3
    }
  },
  {
    head: Point {
      column: 0,
      row: 5
    },
    tail: Point {
      column: 0,
      row: 6
    }
  }
]
should equal
[
  {
    head: {
      column: 2,
      row: 4
    },
    tail: {
      column: 0,
      row: 3
    }
  },
  {
    head: {
      column: 0,
      row: 5
    },
    tail: {
      column: 0,
      row: 6
    }
  },
  {
    head: {
      column: 4,
      row: 11
    },
    tail: {
      column: 0,
      row: 10
    }
  },
  {
    head: {
      column: 0,
      row: 20
    },
    tail: {
      column: 5,
      row: 20
    }
  }
]
      + expected - actual
       [
         {
           "head": {
      -      "column": 4
      -      "row": 11
      +      "column": 2
      +      "row": 4
           }
           "tail": {
             "column": 0
      -      "row": 10
      +      "row": 3
           }
         }
         {
           "head": {
             "column": 0
      -      "row": 20
      +      "row": 5
           }
           "tail": {
      -      "column": 5
      -      "row": 20
      +      "column": 0
      +      "row": 6
           }
         }
         {
           "head": {
      -      "column": 2
      -      "row": 4
      +      "column": 4
      +      "row": 11
           }
           "tail": {
             "column": 0
      -      "row": 3
      +      "row": 10
           }
         }
         {
           "head": {
             "column": 0
      -      "row": 5
      +      "row": 20
           }
[...]

PR #515, opened 2020-03-20
Failing build on Dev channel

darangi:
### Description of the Change
[The dev channel on the CI currently fails](https://github.com/atom/teletype/issues/514), this is because it tries to build on atom's master branch which uses electron 5. Electron 5's API has a slight change in the values used in making the tests assertions. This PR helps to conditionally implement the assertions based on the channel the CI is using.
### fixes https://github.com/atom/teletype/issues/514


lkashef:
elegant :) love it 👍

PR #530, opened 2020-04-21
Add avatars below cursors

KoltesDigital:
This is not a PR yet, but rather a first step to ask:

- Is this feature wanted?
- How do you feel about adding a config to enable/disable this feature?
- How the avatars could be bound to the editor area?
- How could it be automatically tested?

### Description of the Change

This features adds the site avatar below its cursor. To do so, it uses `TextEditor.decorateMarkerLayer` with type `overlay`.

The fact that it's added into the existing *markerLayersBySite* system has the benefit that the suppression on site disconnection is already handled.

### Alternate Designs

Other avatars are displayed in React elements, but it was more complicated to follow the cursor. I believe it perfectly fits as a use case of overlay decoration.

### Benefits

The participants follow more easily what others do.

### Possible Drawbacks

I saw the avatars are drawn above the bottom bar.

### Verification Process

Only manually tested for now.

### DIFF
--- a/lib/editor-binding.js
+++ b/lib/editor-binding.js
@@ -4,6 +4,7 @@ const path = require('path')
 const {Range, Emitter, Disposable, CompositeDisposable, TextBuffer} = require('atom')
 const {getEditorURI} = require('./uri-helpers')
 const {FollowState} = require('@atom/teletype-client')
+const getAvatarURL = require('./get-avatar-url')
 
 module.exports =
 class EditorBinding {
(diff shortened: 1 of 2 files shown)


sadick254:
This could make the UI too cluttered. I would suggest you look into teletype and confirm if your suggestion can be a package. 
Thanks for the interest. We might review this when the idea is fully fleshed out.

KoltesDigital:
You can see what it would look like: https://www.twitch.tv/videos/711181860

Agreed, I haven't though about the teletype provided service. I'll have a look at it.

PR #534, opened 2020-05-07
Fixes failing CI build

darangi:


lkashef:
Just clarifying that these are fails caused by Electron API changes, caused by upgrading to v5 and it will happen again when the Electron Upgrade PR gets to stable.

This looks fine but in the future, we should maybe come up with a more elegant solution, rather than updating the tests with every release though 🙈

